PR #2, opened 2021-03-31
add initial ORM files

moonpatroller:


jd2rogers2 (approved):
started a thread in slack where we can discuss more but idk if we'll need the views
everything else looks great and those are easily enough removed later so 💯 💯 💯

PR #4, opened 2021-04-27
auth for endpoints

jd2rogers2:
create login auth strategy
follows 1st half of https://docs.nestjs.com/security/authentication
i've named it "login" rather than "local" because i feel it's more descriptive of what it does. "local" doesn't tell me anything

so i've added an `/auth/test` endpoint. this is protected by a `UseGuards` decorator which uses our new dependency passport's `'local'` strategy. right now this strategies conventions (uses `username` and `password` properties) don't align with what we have in our DB so we'll have to sync those in a future PR

i've also renamed some `user` stuff to `users` to stick with plural convention

test
1. run server `npm run start:dev`

2. then hit new endpoint
```
curl -X POST http://localhost:3001/auth/test -d '{ "username": "test", "password": "nothing" }' -H "Content-Type: application/json"
```

3. see successful request with response of
`{"message":"ヽ(•‿•)ノ"}`

4. send request without user info
```
curl -X POST http://localhost:3001/auth/test -H "Content-Type: application/json"
```

5. see response for failed request
`{"statusCode":403,"message":"Forbidden resource","error":"Forbidden"}%       `


moonpatroller (commented):
Looks good

moonpatroller (approved):
Looks good

orionsglow (approved):
Approved

PR #9, opened 2021-05-08
Fix for create-react-app dependency conflict

jvendettuoli:
Issue with CRA dependencies conflicting between client node_modules and server node_modules. 

- Changed jest dependency `"jest": "26.6.0"` in both client and server package.json
- Add explicit webpack devDependency `"webpack": "4.44.2"` to server package.json
- Deleted package-lock.json and ran `npm install` in both directories


moonpatroller (approved):
lgtm

PR #13, opened 2021-05-08
adding forms second page

moonpatroller:


jvendettuoli (approved):
Approving but with a question about how we want this to function. One potential issue of showing step 2 "page" through use of the pageNum state change is that if the user wants to go back to the previous "page", they won't able to do so. 

We could get around this by either:
- Adding another button on page 2 to go back (set pageNum to 1). 
- Or we make page 2 its own component with its own route, in which case clicking next will use `history.push(/"signup/nonprofit/page2")` or something like that to store the change in react-router-dom's History, and pressing <- on the browser will actually go back.

First option is easier, but strays from the wireframe design.

PR #14, opened 2021-05-16
NavBar actually navigates

jd2rogers2:
- added ThemeProvider so we don't need to `useTheme` in every component
- add material ui `AppBar` component for header and react-router `Link`s to update url
- added empty components for each of the pages in the navbar
- added react-router `Route` handlers for ☝️ 

can someone add prettier or something 😢 this diff is ugly

1. `cd client && npm run start`
2. http://localhost:3000/contact_us
3. click navbar links
expect page content to change

May-16-2021 12-02-13

### DIFF
--- a/client/src/AboutUs.js
+++ b/client/src/AboutUs.js
@@ -1,0 +1,12 @@
+import React from 'react';
+
+
+function AboutUs() {
+	return (
+		<div>
+      AboutUs
+		</div>
+	);
+}
+
+export default AboutUs;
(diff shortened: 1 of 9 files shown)


jvendettuoli (approved):
Thanks for adding in Theme Provider James. At the time I wasn't sure if people were on board with using Material-UI so I contained it to the page I was working on as an example of what it looks like.

NavBar and empty components look good

Reviewing via GitHub is pretty new to me so I'll leave it up to the more experienced devs as to what might be a nicer diff utility.

PR #15, opened 2021-05-17
Styled Footer and Added empty components and route handlers for footer links

jvendettuoli:
- Styled `<Footer>` to match Figma wireframe
- Added empty components for each of the footer links that didn't exist yet
- Created `<Route>` for each new empty component
- Created `<StyledLink>` component, which takes a `to` prop that sets the href, and is wrapped around text that will be the display text for the link. Currently this component sets the Material-UI Link to use `react-router-dom` Link (imported as `RouterLink` to avoid conflict with Material-UI Link), and add some simple styling that seemed consistent across the wireframes: black color, line on hover, and text aligned to left (only mattered at very small screen sizes).

FooterChanges

### DIFF
--- a/client/src/CookiePolicy.js
+++ b/client/src/CookiePolicy.js
@@ -1,7 +1,7 @@
 import React from 'react';
 
 function CookiePolicy() {
-	return <div>CookiePolicy</div>;
+    return <div>CookiePolicy</div>;
 }
 
 export default CookiePolicy;
--- a/client/src/CookiePolicy.js
+++ b/client/src/CookiePolicy.js
@@ -1,0 +1,7 @@
+import React from 'react';
+
+function CookiePolicy() {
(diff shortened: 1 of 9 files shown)


jd2rogers2 (approved):
couple quick unused lines/vars to remove before landing but otherwise looks good

moonpatroller:
Can you replace tabs with 4 spaces?  Does anyone care about this?

jvendettuoli:
Thanks for catching those unused imports and variables James, I cleaned them out. 

I typically used spaces, but had to reinstall an extension and hadn't realized its default was to convert to tabs on save. Fixed that issue so all these files have 4 spaces. I'm not picky about how we want to format this project. I noticed James has two spaces for his generally. Maybe we can address that quickly in the meeting and set the standard.

PR #16, opened 2021-05-24
Convert React files to TypeScript

jvendettuoli:
Enabling support for TypeScript:
- `npm install --save typescript @types/node @types/react @types/react-dom @types/jest @types/react-router-dom`
- Restart deployment to have tsconfig.json automatically added
- Added `"noImplicitAny": true`  `"noImplicitThis": true` and `"strictNullChecks": true` to tsconfig.json to support Material-UI

Bulk of changes were just changing the file extensions form .js -> .tsx

I then fixed anything that specifically threw an error. This was generally defining types for props, form data, and event handling.

There are two problems I had, that maybe someone has advice for. 
1. Within `SignupNonProfit.tsx`, there are inputs for a checkbox and select. Handling the onChange event for select requires the type `React.ChangeEvent<HTMLSelectElement>` and for other inputs `React.ChangeEvent<HTMLInputElement>` is required.
I attempted to have the event be either one with: `evt: React.ChangeEvent<HTMLInputElement> | React.ChangeEvent<HTMLSelectElement>`
However, the HTMLSelectElement doesn't have a field for checked, so the general onChange handler was throwing an error because we conditionally looked for it if the type of input was checkbox. For now I created two separate handlers, one for select, one for all other inputs.

2. It seems boolean isn't a valid type for input. It expects `string | number | readonly string[] | undefined`. This made it so that I couldn't assign `formData.accept_terms` to expect a boolean, with the default set to `false`. I instead set `accept_terms` to expect a string and set the default to `'false'`. This seems to work, because the checkbox starts unchecked, and when submitted `formData.accept_terms` has a boolean value of `true` or `false` depending on checked or unchecked and no errors are thrown. 

Let me know if there are any other changes I should make. As I mentioned this is my first time working with TypeScript so I may have made some errors I don't know about, or not have used best practices.


client/src/Login.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    interface UserLoginData {
+        email: string;
+        password: string;
+    }
+
+    const initialFormData: UserLoginData = {
+        email: '',
+        password: ''
+    };
```
jd2rogers2, line 54: i would move `UserLoginData` and `initialFormData` out of the FC so they aren't redefined on every rerender
also, IMO `formData` should be broken up into separate pieces of state, because as it is now, the password component would rerender if the email input is changed, but that can be another PR

client/src/Login.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    // Toggle password visibility
+    const handleClickShowPassword = () => {
+        setShowPassword(!showPassword);
+    };
+
+    const handleMouseDownPassword = (evt: React.MouseEvent) => {
+        evt.preventDefault();
+    };
```
jd2rogers2, line 74: is this really needed?

client/src/SignupNonProfit.tsx
```diff
@@ -0,0 +1,191 @@
...
+        tax_exempt_id: '',
+        nonprofit_classification: '',
+        first_name: '',
+        last_name: '',
+        role_or_title: '',
+        email: '',
+        password: '',
+        accept_terms: 'false'
+    };
+
```
jd2rogers2, line 35: move static content out of component

jd2rogers2 (approved):
left a couple nits, they're probs not even on code you changed (just moved) in this PR. if so feel free to ignore and they can be cleaned up at another time
otherwise lgtm

client/src/Login.tsx
jvendettuoli: Agreed, and I have done part of those changes in the Login component I have locally. Figured TypeScript conversion was higher priority so I put that one on hold. I'll have that PR up soon after this.

client/src/Login.tsx
jvendettuoli: No, it was in the Material-UI example when I brought it over. I've got it removed on my login-page branch. Maybe their example had the button as submit, and was submitting the form? Seems to work fine without it on my side. It'll be removed on the login-page PR

client/src/SignupNonProfit.tsx
jvendettuoli: I'll update the PR with this one before merging, since I am not currently working on that component.

PR #18, opened 2021-06-09
more homepage layout and mock content

jd2rogers2:
not perfect but a decent first run i think. if there are styling nits let's leave them until we have more pages ready and we'll do one big cleanup

neh homepage


jvendettuoli (approved):
Agreed no reason to be perfectionists at this point. Great start!

PR #21, opened 2021-06-19
offer page

jd2rogers2:
compare to offer page in mocks https://www.figma.com/file/fYkreARKdHsMh8iPSh7dti/Testing-File?node-id=0%3A1

1. http://localhost:3000/assets
2. click on one of the cards
3. see that you're on the offer page http://localhost:3000/offer/2
4. click on mini img and see that it moves to big img slot
5. click `go back` button, see assets page

not really sure if this page should really be `/assets/2`
but thought we would probs have separate need and offers pages so went with this
also went with primary button color for claim button, not sure why it looks disabled in mocks

Screen Shot 2021-06-19 at 10 33 16 AM

### DIFF
--- a/client/src/Offer.tsx
+++ b/client/src/Offer.tsx
@@ -1,0 +1,127 @@
+import * as React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { useParams } from "react-router-dom";
+// import * as queryString from 'query-string';
+import Paper from '@material-ui/core/Paper';
+import InputBase from '@material-ui/core/InputBase';
+import { NavLink } from 'react-router-dom';
+import SearchIcon from '@material-ui/icons/Search';
+import Button from '@material-ui/core/Button';
+import ArrowBackRoundedIcon from '@material-ui/icons/ArrowBackRounded';
+
+import type { Theme } from '@material-ui/core/styles';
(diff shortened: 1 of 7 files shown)


client/src/Offer.tsx
```diff
@@ -0,0 +1,206 @@
...
+        alignItems: 'center',
+        paddingLeft: '10px',
+    },
+    // TODO i'm just copying these from other files,
+    // either make them a component or make the styles shared objects somewhere
+    iconButton: {
+        padding: 10,
+        '&:hover': {
+            backgroundColor: 'inherit',
+            borderRadius: '10px',
```
jd2rogers2, line 41: i'm just copying a lot of these style objects from other files, we should discuss a better long term solution
just lots of little components?
idk really know material-ui theming but maybe something in there?

jvendettuoli: I think there probably is a better way, but in the past I have saved styles that I found myself reusing all the time into a separate file and imported them into the components, or created styled components. I bet there is a way you can save styles directly to the Theme.

jvendettuoli (approved):
Looks good! We can look a bit deeper into the Material-UI custom themes and see if that is the best way to manage similar styles.

PR #22, opened 2021-06-20
server file restructure

jd2rogers2:
moved all nest related files into `/server`
we had tons of duplicates in `/dist` so ran `npm run build` which deleted old dist and built new one without dups, so i think that seems to be succeeding now


jvendettuoli (approved):
Looks good. Better organization for sure.

PR #25, opened 2021-07-11
Create content for Trust and Safety page

nstjean:
The page is all set up to look like the prototype on Figma. I wasn't sure if there is a specific way to format the font for the project, if the site's stylesheet should do that automatically.

I created SmallDisplayCard to use on this page and others. Text to display is passed in via props.

Code and stylesheet could probably be simplified further, I'm still getting the hang of using React and JSS as opposed to a central stylesheet for a project.

FireShot Capture 278 - Nonprofit Exchange Hub - localhost


jd2rogers2 (approved):
nice work! 💯 
i think you're right about the height and width (or we could just do a small/large prop) being props. but since we only have one use case for now, this is fine

PR #26, opened 2021-07-13
Add encryption for user password. Add method to search for user by email.

jvendettuoli:
Alright, finally a small PR with a simple change! 

Problem: Real user passwords being stored in database is bad practice.
Solution: Hash the passwords on user creation and only store the hashed version. Validate a password on user login.

We are using [bcrypt](https://www.npmjs.com/package/bcrypt) to hash and validate the passwords. Upon user creation, the password is hashed and replaced with the hashed version for insertion into database.  When a user logs in, their submitted password is hashed and compared against the stored hashed password in the database, which is retrieved by searching for a user with a matching email. If they match, the user entity is returned without a password property. If the password is invalid `{
  "status": 401,
  "error": "Invalid password"
}` is returned. If the email is not in our database `{
  "status": 404,
  "error": "Email not found"
}` is returned. 

@jd2rogers2 I know you were talking about having a more generic response to not reveal if a user exists with a given email, but I have left it more specific for now while we are building it to help with trouble shooting. Shouldn't be too hard to make it more generic later. 

Emails are now unique in our database.

Assuming an empty database and not having bcrypt installed.
1.  `npm install` to get `bcrypt` and `@types/bcrypt` installed.
2. You may need to delete `/server/dist` directory and restart server. I did.
3. You'll probably need to drop the existing database schema because the email column of the users table is now unique. The correct version should be automatically created on server restart.
4. Create new ENV variable `BCRYPT_WORK_FACTOR=10`. This is essentially how difficult the hashed password would be to reverse engineer. Higher number means more secure and takes longer to perform. 10 was given in the NestJS Docs, I've seen 12 used pretty often as well. 
5. Create a new user:
```
curl --request POST \
  --url http://localhost:3001/api/users \
  --header 'Content-Type: application/json' \
  --data '{
	"first_name": "TestFirst",
	"last_name": "TestLast",
	"email": "test@test.com",
	"password": "testpassword"
	}'
```
6. Confirm that the password saved is not the real password, but a hashed version that looks like: `$2b$10$s1TpKNUG3LKpysOTw3VrLO9vNnH0vtK6tPsychpfku3YKlIIFus5.`
7. Confirm that logging in with valid credentials gets you an access token:
```
curl --request POST \
  --url http://localhost:3001/api/auth/login \
  --header 'Content-Type: application/json' \
  --data '{
	"email": "test@test.com",
	"password": "testpassword"
}'
```
Should return something like: `{
  "access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2MjYxMzczMjIsImV4cCI6MTYyNjEzNzM4Mn0.EVpkoRXRCO1MIOYt9F5dj5yYrRkYXyDq3BNtj5iIgQU"
}`
8. Confirm that wrong email return error:
```
curl --request POST \
  --url http://localhost:3001/api/auth/login \
  --header 'Content-Type: application/json' \
  --data '{
	"email": "wrong@test.com",
	"password": "testpassword"
}'
```
Returns `{
  "status": 404,
  "error": "Email not found"
}`
9. Confirm that wrong password returns error:
```
curl --request POST \
  --url http://localhost:3001/api/auth/login \
  --header 'Content-Type: application/json' \
  --data '{
	"email": "test@test.com",
	"password": "wrong"
}'
```
Returns `{
  "status": 401,
  "error": "Invalid password"
}`

This PR does not address error handling in the user routes outside of findByEmail. Next PR will be for making routes provide better information beyond a 200 status code if there were no server issues.


server/src/auth/strategies/login.strategy.ts
```diff
@@ -6,15 +6,16 @@ import { AuthService } from '../auth.service';
 
 @Injectable()
 export class LoginStrategy extends PassportStrategy(Strategy) {
-  constructor(private readonly authService: AuthService) {
-    super();
-  }
+    constructor(private readonly authService: AuthService) {
+        // Change the expected validation field from username to email
+        super({ usernameField: 'email' });
```
jd2rogers2, line 11: 👍

server/src/users/users.service.ts
```diff
@@ -1,15 +1,19 @@
...
 @Injectable()
 export class UsersService {
     constructor(@InjectRepository(User) private usersRepository: Repository<User>) {}
-    create(createUserDto: CreateUserDto) {
-        return this.usersRepository.save(createUserDto)
+
+    async create(createUserDto: CreateUserDto) {
+        const hashedPw = await bcrypt.hash(createUserDto.password, parseInt(process.env.BCRYPT_WORK_FACTOR));
+        createUserDto.password = hashedPw;
+        return this.usersRepository.save(createUserDto);
```
jd2rogers2, line 16:
```
jamesrogers@jamesrogpremiseMBP-2 web-app % curl --request POST \
  --url http://localhost:3001/api/users \
  --header 'Content-Type: application/json' \
  --data '{
        "first_name": "TestFirst",
        "last_name": "TestLast",
        "email": "test@test.com",
        "password": "testpassword"
        }'
{"first_name":"TestFirst","last_name":"TestLast","email":"test@test.com","password":"$2b$10$ZFlYKLACsaNmeUFbOxhzr.NAgBl1G3BsnheWYkRFQTVgTzQFtHlDm","id":1}%                                                                                                                                     
jamesrogers@jamesrogpremiseMBP-2 web-app % 
```

create user post request is returning password with it. might wrap this return val (and i suspect the other return vals in this file) with a `getUserSansPw`

```
const getUserSansPw = (user: CreateUserDto): Omit< CreateUserDto, 'password'> => {
  const { password, ...result } = user;
  return result;
};
```
(https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)

jd2rogers2 (approved):
💯 this is awesome, we can probably start connecting FE and BE. this, and the last create user entity PR, are a great POC for adding our other entities. 

i left one comment about the return values of the user service including the password that needs to be addressed, but should be an easy fix so 🚢 

"I know you were talking about having a more generic response to not reveal if a user exists..."
yeah i see both in the real world, not the worst if tell user email doesn't exist i guess. we can punt on this and handle later

"Emails are now unique in our database."
🔥 

"You may need to delete /server/dist directory and restart server. I did."
i think we do something similar to this at work so i think that's expected. looks to me like we should be able to just use `npm run prebuild && npm run build && npm start` as our start script. not too bad but we might consider making a custom script for that
yep, looks like it worked for me. also, i just manually changed `DatabaseConnectionService`'s `dropSchema` to `true` before prebuild + build + start and that worked fine for me. we can use this workflow for now i'd say, there might be a handy flag out there for this somewhere, but we shouldn't be changing our schema that often in the future anyways

server/src/users/users.service.ts
jvendettuoli: Oo Omit, that is very handy. I am planning to have a PR that has the user routes have more informative responses. I could address this issue then.

jd2rogers2: That works too!

PR #27, opened 2021-07-18
serve react from nest RFC

jd2rogers2:
this is just an investigation/proposal at this point
and i haven't done any research on benefits/performance/etc.

this was actually super easy too https://docs.nestjs.com/recipes/serve-static

this would change our infra from having 2 services to only have 1 app
it adds to our start up workflow. to mirror production, you'd have to `build` client side and then do the whole `prebuild && build && start` on the server side
but actually for developing we could still run them separately locally

also:
removed unused user view
removed no longer used appService

testing:
1. `cd server && npm install`
2. `cd ../client && npm run build`
3. `cd ../server && npm run prebuild && npm run build && npm run start`
4. http://localhost:3001 in browser
5. click around and see that page works the same as before


jd2rogers2:
https://www.toptal.com/front-end/client-side-vs-server-side-pre-rendering
https://blog.logrocket.com/why-you-should-render-react-on-the-server-side-a50507163b79/

couple articles from sources i like, seems like the professionals are for it

jvendettuoli:
Tested on my side and works well. If you are using postgres via docker you'll need to run `npm:dev:db`  before starting the server, but that is expected.

jvendettuoli (approved):
Tested on my side and works well. If you are using postgres via docker you'll need to run `npm:dev:db`  before starting the server, but that is expected.

PR #29, opened 2021-08-09
SignupCitizen and Login component updates

jvendettuoli:
Goodness do I wish I had made this PR before I went on vacation.

Overall the changes made were to create reusable subcomponents for the email and password inputs and google/facebook buttons of SignupCitizen and Login components.

Also improved some styles, primarily of SignupCitizen. Definitely still some work to be done, mostly with the sign up button, but I figure it can wait until we have more info from UI/UX crew.

SignupCitizen and Login connected to backend and will successfully create a user and return an access token with the correct inputs. We still need to set restrictions on the password input, currently just required, no min length or anything.

This PR does not handle errors in these cases. That will be for the next PR because it involves changing up response on the backend, and requires answering how we want to handle those situations of failed requests. In the past I've used Redux to handle pending request, and then respond appropriately based on a success or failure. Push to page on success, show errors on failure.

To Test:

- Create user from signup citizen creates a new user in the database and dev console returns 201.
- Trying to create a user without data won't allow submit.
- Logging in with successfully added user email and password returns an access_token, wrong inputs return a 401.

### DIFF
--- a/client/src/SignupCitizen.tsx
+++ b/client/src/SignupCitizen.tsx
@@ -79,12 +79,10 @@ function SignupCitizen() {
             ...fData,
             [name]: name ==='accept_terms' ? checked : value,
         }));
-        console.log(formData);
     };
 
     const handleSubmit = async (evt: React.FormEvent) => {
         evt.preventDefault();
-        console.debug('handleSubmit - formData: ', formData);
         try {
             const res = await fetch('http://localhost:3001/api/users', {
                 method: 'POST',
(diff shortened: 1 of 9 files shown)


client/src/SignupCitizen.tsx
```diff
@@ -1,47 +1,206 @@
...
+        setFormData((fData) => ({
+            ...fData,
+            [name]: name ==='accept_terms' ? checked : value,
+        }));
+        console.log(formData);
+    };
+
+    const handleSubmit = async (evt: React.FormEvent) => {
+        evt.preventDefault();
+        console.debug('handleSubmit - formData: ', formData);
```
jd2rogers2, line 87: remove the logs (1 above here too)

jd2rogers2 (approved):
nit with the logs but otherwise this looks good!!

PR #31, opened 2021-08-11
Create How It Works page

nstjean:
This page has two versions depending on what tab is selected at the top. All of the content is created in variables so that the appropriate content can be displayed.

I've made a sub-component in the page for BulletGrid (the white squares section) and the InstructionGrid (text and images on the top half).

I made a component QuestionList for the clickable drop-down questions since I could see that being used in other places on the site.

FireShot Capture 287 - Nonprofit Exchange Hub - localhost

FireShot Capture 286 - Nonprofit Exchange Hub - localhost

### DIFF
--- a/client/src/DisplayGrids/index.ts
+++ b/client/src/DisplayGrids/index.ts
@@ -1,3 +1,3 @@
-export { BulletGrid } from './BulletGrid';
-export { InstructionGrid } from './InstructionGrid';
-export { GridImages } from './GridImages';
+export { default as BulletGrid } from './BulletGrid';
+export { default as InstructionGrid } from './InstructionGrid';
+export { default as GridImages } from './GridImages';
--- a/client/src/DisplayGrids/index.ts
+++ b/client/src/DisplayGrids/index.ts
@@ -1,0 +1,3 @@
+export { BulletGrid } from './BulletGrid';
+export { InstructionGrid } from './InstructionGrid';
+export { GridImages } from './GridImages';
--- a/client/src/HowItWorks.tsx
+++ b/client/src/HowItWorks.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
 import Typography from '@material-ui/core/Typography';
 import { makeStyles } from '@material-ui/core/styles';
-import { Box, Grid, ButtonBase } from '@material-ui/core';
+import { Box, Grid } from '@material-ui/core';
 import QuestionList from './QuestionList';
+import { BulletGrid, InstructionGrid } from './DisplayGrids';
 
 import type { Theme } from '@material-ui/core/styles';
 
@@ -132,8 +133,6 @@ const useStyles = makeStyles((theme: Theme) => ({
             paddingBottom: '0!important',
         },
     },
-    grid: {
-    },
     button: {
         fontSize: '2rem',
         margin: '20px 0 0 0',
@@ -145,14 +144,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     selected: {
         borderBottom: '11px solid #C4C4C4',
     },
-    nonprofitContainer: {
-        // height: 200,
-        // backgroundColor: 'blue',
-    },
-    citizenContainer: {
-        // height: 200,
-        // backgroundColor: 'red',
-    },
     questions: {
         alignSelf: 'flex-end',
         width: '100%',
@@ -163,9 +154,11 @@ const useStyles = makeStyles((theme: Theme) => ({
 function HowItWorks() {
     const classes = useStyles();
 
-    const [ tabSelected, setTabSelected ] = React.useState("nonprofit");
+    type Tab = 'nonprofit' | 'citizen';
 
-    const handleClickTab = (tabName: string) => {
+    const [ tabSelected, setTabSelected ] = React.useState<Tab>("nonprofit");
+
+    const handleClickTab = (tabName: Tab) => {
         setTabSelected(tabName);
     };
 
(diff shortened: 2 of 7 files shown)


client/src/HowItWorks.tsx
```diff
@@ -1,7 +1,341 @@
...
+        borderBottom: '11px solid #C4C4C4',
+    },
+    nonprofitContainer: {
+        // height: 200,
+        // backgroundColor: 'blue',
+    },
+    citizenContainer: {
+        // height: 200,
+        // backgroundColor: 'red',
+    },
```
jd2rogers2, line 155: `grid`, `nonprofitContainer`, `citizenContainer` remove if unused

client/src/HowItWorks.tsx
```diff
@@ -1,7 +1,341 @@
...
+        width: '100%',
+        maxWidth: '700px',
+    },
+}));
 
 function HowItWorks() {
-    return <div>HowItWorks</div>;
+    const classes = useStyles();
+
+    const [ tabSelected, setTabSelected ] = React.useState("nonprofit");
```
jd2rogers2, line 166: like this, typescript will set this state as an `any` type
we can take advantage of enums here to enforce that it's only 1 of the 2 possibilities
and then typescript will catch any errors therefore giving us more confidence
so `React.useState<'nonprofit' | 'citizen'>("nonprofit");`

client/src/HowItWorks.tsx
```diff
@@ -1,7 +1,341 @@
...
+        backgroundColor: 'white',
+        marginRight: '20px',
+    },
+});
+
+type BulletProps = {
+    list: string[],
+};
+
+function BulletGrid(props: BulletProps) {
```
jd2rogers2, line 313: yeah i guess let's pop these into separate files, this one is getting lengthy
checkout the file structure of `AssetsView` from my branch in this PR https://github.com/Nonprofit-Exchange-Hub/web-app/pull/32/files

client/src/HowItWorks.tsx
```diff
@@ -1,7 +1,341 @@
...
+    instructionList: { title: string, body: string, image: string }[],
+};
+
+function InstructionGrid(props: InstructionProps) {
+    const classes = instructionStyles(props);
+    var row = 0;
+
+    return (
+        <Grid container justify='space-between'>
+            {props.instructionList.map((instructionItem) => {
```
jd2rogers2, line 258: use the map's index (`.map(instructionItem, index) =>`) instead of `row`

client/src/HowItWorks.tsx
```diff
@@ -1,7 +1,341 @@
...
+                    <Grid item sm={6} xs={12} className={`${classes.gridBox}`}>
+                        <Typography variant="h3" component="h3" align="left">
+                            {instructionItem.title}
+                        </Typography>
+                        <Typography variant="body1" component="div" align="left" className={`${classes.headerText}`}>
+                            {instructionItem.body}
+                        </Typography>
+                    </Grid>
+                );
+                var image = (
```
jd2rogers2, line 270: i would say pull `text` and `image` out to be separate components as well. these can live in the same file as `InstructionGrid` since they're pretty small

client/src/QuestionList.tsx
```diff
@@ -0,0 +1,88 @@
...
+    questionList: {
+        width: '100%',
+    },
+});
+
+type QuestionListProps = {
+    questionList: { question: string, answer: string }[],
+};
+
+function QuestionList(props: QuestionListProps) {
```
jd2rogers2, line 23: aaaayee bummer
sorry, i should've caught this in the designs earlier and said something to the team
this should use the same component as the homepage
we wouldn't want 2 accordion-esque components around, would make our page look inconsistent
checkout https://github.com/Nonprofit-Exchange-Hub/web-app/blob/9e0f96dfcb7bc0031e5750523b47f5af46f2943e/client/src/Home.tsx#L194

jd2rogers2 (commented):
everything looks pretty good!
couple changes i would make but should be a quick approval on the next round

client/src/HowItWorks.tsx
nstjean: Oh that's great!

client/src/HowItWorks.tsx
nstjean: Sure!

client/src/HowItWorks.tsx
nstjean: Thanks!

client/src/HowItWorks.tsx
nstjean: Sounds good

client/src/QuestionList.tsx
nstjean: No worries! It was good for me to figure it out, but it's easy to change.

nstjean:
I've made all the requested changes except the questions drop-down. I think it makes sense to make another pull request for updating How It Works and the home page to use the same module for the questions drop-down, and style them consistently.

client/src/DisplayGrids/index.ts
```diff
@@ -0,0 +1,3 @@
+export { BulletGrid } from './BulletGrid';
+export { InstructionGrid } from './InstructionGrid';
```
jd2rogers2, line 2: in `InstructionGrid` you can do `export default InstructionGrid;`
and here you could do `export { default as InstructionGrid } from './InstructionGrid';`

jd2rogers2 (approved):
one nit. and yes i agree, the accordion consolidation can be a separate PR
nice work 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳

PR #32, opened 2021-08-12
message inbox view, user context

jd2rogers2:
[mocks](https://www.figma.com/file/fYkreARKdHsMh8iPSh7dti/GC-prototype-v1?node-id=0%3A453)
i didn't spend much time on getting the styling exactly right since these are just wireframes and will likely change
mostly focused on functionality and some page layout

so this PR adds:
- user context provider (meaning we can now get user info in pretty much any view and display their info, or show different UIs based on if user is logged in, user type, etc.)
- a ternary in the `Header` component to display a user account button/drop down, when logged in, instead of the "sign up"/"sign in" buttons
- an `/inbox` page with clickable message threads, but you cannot send messages
- an abstraction which i called `SubHeader` and left a TODO to use it in the other places

view with empty inbox
Screen Shot 2021-09-07 at 8 14 14 AM

repro steps
1. http://localhost:3000/inbox
2. see that you are redirect to home because you are not logged in
3. log in with the user that has an ID of `3` in your DB
4. click on the user account icon in the top right, select inbox from drop down
5. see that you have messages in your inbox and you have an open thread
user login


client/src/SubHeader.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
```
jd2rogers2, line 1: so yeah, i didn't really write any of the code from this file, mostly just copied from where we have other sub headers (usually with a 'back' and search input)
however we have inconsistent sub headers so will be bringing this to the design team to hopefully clean up
will reuse this component in all the other places in a separate PR

client/src/Header.tsx
```diff
@@ -44,6 +50,18 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 function Header() {
     const classes = useStyles();
+    const [user] = React.useContext(UserContext);
+    const history = useHistory();
+
+    const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
jd2rogers2, line 56: material-ui throws type error when i try to use ref :(

nstjean (commented):
I'm getting an error when I try to start the server. Is there something I need to do differently?

FireShot Capture 290 - Nonprofit Exchange Hub - localhost

jd2rogers2:
> I'm getting an error when I try to start the server. Is there something I need to do differently?
> 
> FireShot Capture 290 - Nonprofit Exchange Hub - localhost

@nstjean di you...
`cd client`
`npm install`
?

jvendettuoli:
Holy smokes I just ran into a big issue trying to checkout this branch. Turns out my version of npm was too old? Running `npm install` to get jsonwebtokens somehow broke material-ui types. But only on this branch, not on main, where the material-ui type changes were made by Natalie to handle the custom Theme. No idea why. That just took me a  couple hours of troubleshooting.

With it finally working I went through your steps and they work.

jvendettuoli (approved):
Approved - Comments on main PR

PR #33, opened 2021-08-16
Change QuestionList module to use Accordion

nstjean:
Updated QuestionList to use Accordion instead of List
Used QuestionList on Home and How It Works

FireShot Capture 288 - Nonprofit Exchange Hub - localhost

FireShot Capture 289 - Nonprofit Exchange Hub - localhost

### DIFF
--- a/client/src/QuestionList.tsx
+++ b/client/src/QuestionList.tsx
@@ -35,32 +35,35 @@ function QuestionList(props: QuestionListProps) {
 
 const questionStyles = makeStyles<Theme, QuestionProps> ({
     questionWrapper: {
-        boxShadow: 'none!important',
+        boxShadow: 'none',
         borderBottom: '1px solid black',
-        '& span': {
+        '& .MuiTypography-body1': {
             fontSize: '1.2rem',
         },
-        '&.Mui-expanded': {
-            margin: '0',
+        '&.MuiAccordion-root:before': {
+            backgroundColor: 'transparent',
         },
-        '& .MuiAccordionSummary-content': {
-            '& p': {
-                fontWeight: 'bold',
+        // Overrides to remove extra margin when answer is displayed
+        '&.Mui-expanded': {     
+            margin: '0',    
+            '& .MuiAccordionSummary-root': {
+                minHeight: '52px',
+                '& .MuiAccordionSummary-content': {
+                    margin: '12px 0',
+                },
             },
         },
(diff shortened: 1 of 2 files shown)


client/src/QuestionList.tsx
```diff
@@ -37,17 +35,33 @@ function QuestionList(props: QuestionListProps) {
 
 const questionStyles = makeStyles<Theme, QuestionProps> ({
     questionWrapper: {
+        boxShadow: 'none!important',
```
jd2rogers2, line 38: just tried it out, i don't think `important` is needed here

client/src/QuestionList.tsx
```diff
@@ -37,17 +35,33 @@ function QuestionList(props: QuestionListProps) {
...
+            minHeight: '48px',
+        },
+        '& .MuiButtonBase-root:hover': {
+            backgroundColor: 'rgba(240, 240, 240, 0.5)',
+            
+        },
+        '&.MuiAccordion-root:before': {
+            transition: 'none',
+            backgroundColor: 'transparent',
+        },
```
jd2rogers2, line 64: hmmm
i'm thinking there should be a separate class on the summary and some of these would go in a separate root level key of this object that is paired with that
do the &expanded bits apply to it only when expanded or does it look ok when it's not expanded if we put those on the root?
either way, don't spend too much time looking into these, nbd

jd2rogers2 (approved):
couple small nits before deploy. but overall, nice abstraction! 💯

nstjean:
I'll take a look at those. I seriously appreciate you pointing out details, it helps me improve for next time. :)

client/src/QuestionList.tsx
nstjean: the expanded bits only apply when it's expanded. It's a bit annoying because some of these styles I'm just trying to get rid of what comes with Accordion.

nstjean: I re-organized it, see what you think. It's not as clean as I would like it due to those overrides but I didn't want the text bouncing up and down when people click on questions.

jd2rogers2: yeah it looks fine
originally i was in the train of thought that it'd be more readable to have the logic in JS, but i think that would actually include managing extra state in a onHover action. not ideal
so this might be better anyways
i say we ship now and worry about it later, progress > perfection

PR #35, opened 2021-08-24
Create custom theme

nstjean:
Created a Theme for the project! This allowed me to set a standard font and colors and move global styles out of components. This was fun, I look forward to playing more with it.

There are some pretty interesting changes coming in Material UI v5 (currently in alpha), including setting custom variants for existing components (https://next.material-ui.com/customization/theme-components/#adding-new-component-variants). The theme file will probably have to be re-written once we upgrade to v5.

I made some notes and posted them here about how to set up and use the Theme: https://www.notion.so/nstjean/Global-Theming-393d9a09037e45178740d89a03aafce4

### DIFF
--- a/client/src/AboutUs.tsx
+++ b/client/src/AboutUs.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import Typography from '@material-ui/core/Typography';
 import { makeStyles } from '@material-ui/core/styles';
-import { Box, Grid } from '@material-ui/core';
+import { Box, Grid, Container } from '@material-ui/core';
 
 import SmallDisplayCard from './SmallDisplayCard';
 import { GridImages } from './DisplayGrids';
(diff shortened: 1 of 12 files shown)


client/src/AboutUs.tsx
```diff
@@ -10,47 +10,51 @@ import type { Theme } from '@material-ui/core/styles';
...
-        display: 'flex',
-        flexDirection: 'column',
-        alignItems: 'center',
-        justifyContent: 'center',
-        paddingTop: '50px',
-        paddingBottom: '50px',
-        paddingLeft: '20px',
-        paddingRight: '20px',
-        fontSize: '1.4rem',
-    },
```
jd2rogers2, line 35: nice! this all got moved to the theme?!

client/src/Theme.tsx
```diff
@@ -0,0 +1,80 @@
...
+            justifyContent: 'center',
+        },
+    },
+    MuiTypography: {
+        h3: {
+            width: '100%',
+            marginBottom: '35px',
+        },
+    },
+};
```
jd2rogers2, line 78: interesting that these `overrides` aren't just in the `createMuiTheme` arg
would be great to have you walk the rest of us through the architecture and the whys at the next eng mtg

client/src/TrustAndSafety.tsx
```diff
@@ -97,97 +60,103 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+                                    bodyText={loremIpsum.slice(0,56)}
+                                />
+                            </Grid>
+                            <Grid item xs={12} sm={4} container direction="column" alignItems="center">
+                                <SmallDisplayCard
+                                    headerText="Privacy"
+                                    bodyText={loremIpsum.slice(0,56)}
+                                />
+                            </Grid>
+                        </Grid>
```
jd2rogers2, line 108: it's fine for this PR but it'd be good to get in the habit of keeping your cleanup out of a bigger PR like this
it's best to have each PR have a single responsibility

client/src/types/material-ui.d.ts
```diff
@@ -0,0 +1,15 @@
...
+      maxContentWidth: string;
+    }
+  }
+  // allow configuration using `createMuiTheme`
+  interface ThemeOptions {
+    custom?: {
+        maxContentWidth?: string;
+    };
+  }
+}
```
jd2rogers2, line 15: hmm, i feel like this shouldn't be necessary
what made you do this and was there a resource that told you to do it this way?

client/src/HowItWorks.tsx
```diff
@@ -142,7 +114,7 @@ const useStyles = makeStyles((theme: Theme) => ({
         display: 'none',
     },
     selected: {
-        borderBottom: '11px solid #C4C4C4',
+        borderBottom: `11px solid ${theme.palette.primary.main}`,
```
jd2rogers2, line 117: 🔥

client/src/Home.tsx
```diff
@@ -89,11 +89,12 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-        fontStyle: 'normal',
-        fontWeight: 'normal',
-        fontSize: '30px',
-        lineHeight: '45px',
+        padding: '0 0 0 5%',
+        // fontFamily: 'Poppins',
+        // fontStyle: 'normal',
+        // fontWeight: 'normal',
+        // fontSize: '30px',
+        // lineHeight: '45px',
```
jd2rogers2, line 97: remove commented or leave a note in here as to why you're keeping it in

client/src/App.tsx
```diff
@@ -9,7 +10,7 @@ import Routes from './Routes';
 import './App.css';
 
 function App(): JSX.Element {
-    const theme = useTheme();
+    // const theme = useTheme();
```
jd2rogers2, line 13: 🪓

client/src/App.css
```diff
@@ -35,4 +35,4 @@
   to {
     transform: rotate(360deg);
   }
-}
+} */
```
jd2rogers2, line 38: so what becomes of this file?
find anything online that says we should keep or remove?
if we're keeping for now, let's leave a `// TODO find out if we still need this or can remove`

jd2rogers2 (approved):
nice work!!
left some small ones in there and one about the typing file you had build that would be good to sort out, but could also be done in a follow on PR
up to you but let's not let this block us or slow us down

client/src/Theme.tsx
nstjean: Right? It's weird the way they have it right now. I'm looking forward to v5, from what I can see they combine it all into `createMuiTheme`

client/src/TrustAndSafety.tsx
nstjean: Ok, got it

client/src/types/material-ui.d.ts
nstjean: I believe it was here: https://stackoverflow.com/questions/61120760/how-to-extend-material-ui-theme-with-typescript
In order to create a custom variable it has to be declared in Typescript or it will throw an error.

client/src/AboutUs.tsx
nstjean: Yes! :)

client/src/App.css
nstjean: Ok. I think we can delete it since it's not used, but I wanted to be cautious

client/src/types/material-ui.d.ts
nstjean: Oh here it is: https://www.mi4t.me/blog/howt-to-extend-material-ui-theme/

client/src/AboutUs.tsx
nstjean: Yes! :)

client/src/App.css
nstjean: I think we can remove it, but I didn't want to be hasty. Adding a note is good.

client/src/types/material-ui.d.ts
nstjean: TypeScript complains if you don't declare the variable like this

PR #43, opened 2021-10-14
create Assets module, controller, and service worker

Sunset05:
This PR adds the logic on the server side to handle API logic and routing for the Asset model in the database.


server/src/Assets/assets.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  async findOne(id: number) {
+    const asset = await this.assetsRepository.findOne({ id });
+    if (!asset) {
+      throw new HttpException(
+        { status: HttpStatus.NOT_FOUND, error: 'Asset id not found' },
+        HttpStatus.NOT_FOUND,
+      );
+    }
```
jd2rogers2, line 29: i think this should be handled in the controller if possible
so return value here might really be something like `: Promise<Asset | null>`

jd2rogers2: controllers generally handle errors so that they're all in one place (controllers could be hitting multiple services)
also, we'll want to add these errors to the request object so that it can automagically be put into our hosting service's (i.e. AWS) logs

server/src/Assets/dto/update-asset.dto.ts
```diff
@@ -0,0 +1,4 @@
+import { PartialType } from '@nestjs/mapped-types';
+import { CreateAssetDto } from './create-asset.dto';
+
+export class UpdateAssetDto extends PartialType(CreateAssetDto) {}
```
jd2rogers2, line 4: @jvendettuoli this follows the same pattern you used with `UpdateUserDto`
do you remember why you had a separate DTO here rather than just reuse the `CreateUserDto` (which i think we would have all of the same properties of)

jd2rogers2: @Sunset05 give it 5 minutes of work to see if we can just reuse the createDto

server/src/Assets/entities/asset.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+@Entity('assets')
+export class Asset {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column('text')
+  title: string;
+
+  @Column('text')
+  descripton: string;
```
jd2rogers2, line 13: spelling

server/src/Assets/entities/asset.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+  // type: string;
+
+  // @Column('text')
+  // image_url: string;
+
+  // @Column('text')
+  // address: string;
+
+  @ManyToOne(() => User, (user) => user.assets)
+  poster_id: User;
```
jd2rogers2, line 25: looks like somewhere along the line (maybe typeorm 🤷 ) the `id` suffix gets added for us, so let's have this just be `poster`
Screen Shot 2021-10-20 at 6 19 25 PM

Sunset05:
I have recently updated my PR to cover all the comments from the last code review. I have attached images from Postman testing my routes and covered all CRUD actions. I also tested POST requests with bad data to see what would happen and they were rejected by the server. 

Typing added to every function in assets.controller.ts (create(), findAll(), ...etc). 
 
 
POST-with-correct-condition

POST-with-bad-data

PATCH-one

GET-one

GET-all-assets

DELETE-one

final-GET-after-delete

Sunset05:
If you want to test these routes you can send requests to

GET - http://localhost:3001/api/assets
POST - http://localhost:3001/api/assets
PATCH and DELETE - http://localhost:3001/api/assets/:id

Sunset05:
with most recent commit, validations were added to show more meaningful messages from the server when requests are not properly formatted

ex) asking for an asset that doesn't exist
image

ex) PATCH request with invalid format
image

ex) Deleting an asset that does not exist
image

Sunset05:
last commit. upon reviewing  my code I realized that the response from the server sometimes had a key value pair of error, and other times message. Made the response object always contain a k:v pair with message.

PR #69, opened 2021-11-02
claim button

paaaarv:
-added organization to asset to differentiate between individuals/nonprofit accounts 

- added simple snackbar component 

-claim button sucessfully shows snackbar when clicked


client/src/Asset.tsx
```diff
@@ -173,29 +217,29 @@ function Asset(): JSX.Element {
...
-                        Posted By {asset.postedBy}
-                    </Typography>
+                    {postedFrom()}
                     <Typography className={classes.subText} variant="subtitle1">
                         <RoomOutlined />{asset.location}
                     </Typography>
                     <Typography className={classes.subText} variant="subtitle1">
                         <TodayOutlined />{asset.datePosted}
                     </Typography>
+                    {aboutInfo()}
```
jd2rogers2, line 233: i would've said just change to the snippet below. but i think i see why you're trying to use these subrenders, because this render is getting lengthy?
if that's the case then i would say we break this into a data-wrapper and display-component pattern. which is this: we have this `Asset` component that handles all of the state and logic etc. and the only styling it handles is the spacing between the 2 children, which would be those display-components aka something like `AssetViewLeftPanel` and `AssetViewRightPanel`, and these kiddos are really dumb components, they shouldn't have anything but a render/return in the function if possible
so yeah this seems like a bigger change so maybe do the below snippet for this PR and then we can clean it up in a fast follow PR? lmk what you think
```
{asset.organization && (
                  <div>
                    <Typography className={classes.linkedText} variant="subtitle1">
                    About the Nonprofit
                    </Typography>
                    <Typography className={classes.subText} variant="subtitle1">
                    {asset.description}
                    </Typography>
                </div>
)}
```

client/src/Asset.tsx
```diff
@@ -173,29 +217,29 @@ function Asset(): JSX.Element {
...
+                    <Typography className={classes.heading} variant="h3">
                         {asset.title}
                     </Typography>
                     <Typography className={classes.subText} variant="subtitle1">
                         {asset.categories.join(', ')}
                     </Typography>
-                    <Typography className={classes.subText} variant="subtitle1">
-                        Posted By {asset.postedBy}
-                    </Typography>
+                    {postedFrom()}
```
jd2rogers2, line 226:
```
                <Typography className={asset.organization ? classes.linkedText : classes.subText} variant="subtitle1">
                    {asset.organization ? ''  : 'Posted By '}{asset.postedBy.firstName}
                </Typography>
```

client/src/SimpleSnackbar.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { withStyles } from '@material-ui/core/styles';
+import Button from '@material-ui/core/Button';
+import Snackbar from '@material-ui/core/Snackbar';
+import IconButton from '@material-ui/core/IconButton';
+import CloseIcon from '@material-ui/icons/Close';
+import CheckCircleIcon from '@material-ui/icons/CheckCircle'
+import {makeStyles, Theme} from '@material-ui/core/styles';
```
jd2rogers2, line 8: theme is a type not a value so it's conventional to break this up into...
```
import {makeStyles} from '@material-ui/core/styles';
import type {Theme} from '@material-ui/core/styles';
```

client/src/SimpleSnackbar.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const useStyles=makeStyles((theme:Theme)=>({
+    root:{
+        width: '100%',
+    },
+}))
+
+export default function SimpleSnackbar(){
+    const classes=useStyles()
+    const [open, setOpen]=React.useState(true)
```
jd2rogers2, line 19: conventional to prefix with an `is` or a `b` (for boolean) for boolean vars
because otherwise an `open` prop could be misconstrued as a function (open is also a verb)
also do the typing thing here too `useState<boolean>(`

client/src/SimpleSnackbar.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    const handleClick = () =>{
+        setOpen(true)
+    }
+    const handleClose = (event?: React.SyntheticEvent, reason?: string) =>{
+        if(reason=="clickaway"){
```
jd2rogers2, line 25: this if block is doing nothing, did you mean to `return` in there so that we don't hit the `setIsOpen`?
if not let's remove

client/src/SimpleSnackbar.tsx
```diff
@@ -0,0 +1,56 @@
...
+        }
+        setOpen(false)
+    }
+
+
+        return(
+            <div className={classes.root}>
+                
+                <Snackbar
+                anchorOrigin={{vertical: "bottom", horizontal: "right"}}
```
jd2rogers2, line 36: indentation, and remove white space above

client/src/SimpleSnackbar.tsx
jd2rogers2: also triple equals always plz

jd2rogers2 (commented):
this looks really good! mostly small fixes needed, and i like your intuition to try to break up the larger render

not going to approve though because i see one thing that's really funky. see in gif below that a new image pops up in the image thumbnails when i click on the `claim` button. do a little investigation here and clean up the small things and then the next round should be a quick approve
snackbar bug

jd2rogers2:
one more thing, i'm seeing some warnings in the react terminal process, let's try to fix any that were caused by this PR
```
src/SimpleSnackbar.tsx
  Line 2:10:   'withStyles' is defined but never used            @typescript-eslint/no-unused-vars
  Line 3:8:    'Button' is defined but never used                @typescript-eslint/no-unused-vars
  Line 7:8:    'CheckCircleIcon' is defined but never used       @typescript-eslint/no-unused-vars
  Line 21:11:  'handleClick' is assigned a value but never used  @typescript-eslint/no-unused-vars
  Line 25:18:  Expected '===' and instead saw '=='               eqeqeq
  ```

File: client/src/SimpleSnackbar.tsx [hunk 7]
```diff
@@ -0,0 +1,41 @@
...
+const useStyles=makeStyles((theme:Theme)=>({
+    root:{
+        width: '100%',
+    },
+}))
+
+export default function SimpleSnackbar(){
+    const classes=useStyles()
+    const [isOpen, isSetOpen]=React.useState<boolean>(true)
+    const handleClose = (event?: React.SyntheticEvent, reason?: string) =>{
```
jd2rogers2, line 17: let's remove unused args

client/src/SimpleSnackbar.tsx
```diff
@@ -0,0 +1,41 @@
...
+                    <IconButton
+                      key="close"
+                      aria-label="Close"
+                      color="inherit"
+                      onClick={handleClose}
+                    >
+                      <CloseIcon />
+                    </IconButton>,
+                    ]}
+                    message={<span id="message-id">You claimed this! </span>}
```
jd2rogers2, line 37: remove trailing space

client/src/Asset.tsx
```diff
@@ -156,46 +200,38 @@ function Asset(): JSX.Element {
...
                     </Typography>
                     <Typography className={classes.subText} variant="subtitle1">
                         <TodayOutlined />{asset.datePosted}
                     </Typography>
+                    {aboutInfo()}
                     <Button
                         color="primary"
                         variant="contained"
-                        onClick={handleClaim}
+                        onClick={(e:any)=>handleClaim(e)}
```
jd2rogers2, line 229: can be just `handleClaim` now

client/src/Asset.tsx
```diff
@@ -156,46 +200,38 @@ function Asset(): JSX.Element {
...
+                    <Typography className={classes.heading} variant="h3">
                         {asset.title}
                     </Typography>
                     <Typography className={classes.subText} variant="subtitle1">
                         {asset.categories.join(', ')}
                     </Typography>
-                    <Typography className={classes.subText} variant="subtitle1">
-                        Posted By {asset.postedBy}
-                    </Typography>
+                    {asset.organization ? "": "Posted By:"}{asset.postedBy.firstName}
```
jd2rogers2, line 218: why remove the `Typography` here?
would be good to keep it in even if you change the styling or variant

client/src/Asset.tsx
```diff
@@ -116,11 +127,18 @@ function Asset(): JSX.Element {
...
+
+
 
-    const handleClaim = () => {
+
+    const handleClaim = (e:any) => {
+
+        if(e.currentTarget.type==="button"){
+            setShowSnackbar(true)
+    }
```
jd2rogers2, line 141:
```
    const handleClaim = (e: React.HtmlButtonElement) => {
            setShowSnackbar(true);
    }
```
we shouldn't need the if, since the only place we put this handler function is on the button element
unless maybe you were trying to appease typescript? changing the arg type should do the trick

client/src/Asset.tsx
```diff
@@ -133,6 +151,32 @@ function Asset(): JSX.Element {
 
     const bigImg = asset.imgUrls[selectedImgInd];
 
+    const showMiniImgs =
+        asset.imgUrls.map((imgUrl, ind) => {
+          if (ind !== selectedImgInd) {
+            return <img
+                key={imgUrl + `_${ind}`}
```
jd2rogers2, line 158: sorry, i'm going to change my mind again
in the future i'm planning for us, we should never have duplicate image urls in this array
so let's keep the key as just `imgUrl` and let's make the fix at the source. aka remove the duplicate image from the dummy data source
if that doesn't work for any reason just keep this in here and we can figure out down the line

jd2rogers2 (approved):
only a couple nits so go ahead and 🚀 after those are fixed
🥳 🍰 🥳 🍰🥳 🍰🥳 🍰 🥳 🍰🥳 🍰🥳 🍰 🥳 🍰🥳 🍰

PR #76, opened 2021-11-21
update readme

jd2rogers2:
- cleanup
- adds helpful tips for common snags in our set up steps


zoeblairfriedman:
Hey James, I created my own branch based off of yours, and then pushed mine to: "origin HEAD:jrogers/update_readme_setup". Hopefully that was the move! 

I also need to figure out how to preview the formatting of it. I'm a little worried that it doesn't look very clear.

jd2rogers2:
@zoeblairfriedman i just formatted it a bit more, but feel free to take it further if you want (i don't think you need to)
but previewing is possible right in vscodw
Screen Shot 2021-11-22 at 3 41 47 PM 2

or you could've viewed our repo's homepage (shows readme by default) and switched to this branch
https://github.com/Nonprofit-Exchange-Hub/web-app/tree/jrogers/update_readme_setup

as for git workflow, i guess what you did worked fine but even simpler, you can just work right in my branch on your local and then push into my remote branch (fine as long as you're not deleting any of my commits because i'll always be able to see what you changed)

README.md
```diff
@@ -34,11 +29,19 @@ Primary tech stack: PostgreSQL, Nest, and React
     DATABASE_DB=test_db
     BCRYPT_WORK_FACTOR=10   
 ```
+6. In that `.env` file we'll now customize some of those values
+    * If you are using the non-dockerized version of postgres, change the `DATABASE_PORT` value to `5432`
+
+TODO add helpful commands for:
+finding postgres user
+finding postgres db name
+changing postgres user's password via sudo or su (superuser)
```
jd2rogers2, line 38: @Sunset05 this is where you come in 😄

Sunset05: I'll get to it!

zoeblairfriedman:
@jd2rogers2 Ah, yes. Thanks for that screenshot. I forgot about that feature in VSCode! Super helpful. Also I see how you've formatted the code sections now. This looks a lot better, thank you. Just going to make one slight header adjustment but I know Keith is on the rest of it!

PR #78, opened 2021-11-23
skip broken app test and fix a bad import that was breaking the others

jd2rogers2:
we've neglected our tests :(
but at least running them did catch one error that we made 💯 
i'm creating an issue to build out some of these tests
in the meantime:

run tests, and see they pass
`cd server && npm run test`
Screen Shot 2021-11-23 at 11 58 35 AM


yndajas (approved):
Works for me (checked out your branch, ran `npm run test` and got 1 skipped, 5 passed as per your screenshot)

Had to run `npm install` again to fix error `sh: 1: jest: not found`

PR #46, opened 2021-10-20
Messages

yndajas:
PR adds messages entity/controller etc.

Steps to test (adapted from #24):

1. create user
```
curl --request POST \
 --url http://localhost:3001/api/users \
 --header 'Content-Type: application/json' \
 --data '{
   "first_name": "TestFirst",
   "last_name": "TestLast",
   "email": "test@test.com",
   "password": "testpassword"
   }'
```

Returns: `returns {"first_name":"TestFirst","last_name":"TestLast","email":"test@test.com","password":"testpassword","id":1}`

2. log in and note token
```
curl --request POST \
  --url http://localhost:3001/api/auth/login \
  --header 'Content-Type: application/json' \
  --data '{
	"email": "test@test.com",
	"password": "testpassword"
}'
```

Returns (value changes each time): `{"access_token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MiwiZW1haWwiOiJ0ZXN0QHRlc3QuY29tIiwiaWF0IjoxNjM2NTEwNjAyLCJleHAiOjE2MzY1MTA2NjJ9.7TU24FHgB-j3yn3UGE05pTWnDS8Mm_nu9htiOk0P01I"}`

3. send a create message post request, replacing the token in the header with the token returned after the previous step
```
curl --request POST \
--url http://localhost:3001/api/messages \
-H 'Content-Type: application/json' \
-H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MiwiZW1haWwiOiJ0ZXN0QHRlc3QuY29tIiwiaWF0IjoxNjM2NTEwNjAyLCJleHAiOjE2MzY1MTA2NjJ9.7TU24FHgB-j3yn3UGE05pTWnDS8Mm_nu9htiOk0P01I' \
--data '{ "text": "this is a test message", "transaction_id": "1" }'
```

Returns: `{"text":"this is a test message","transaction_id":"1","id":9,"created_date":"2021-11-10T02:17:03.570Z"}`

PR #80, opened 2021-11-23
import Repository and DeleteResult as types

yndajas:
- Import `Repository` and `DeleteResult` from `typeorm` as types in asset, message and user services
- Reorder imports in relevant files so that framework and project-internal imports are grouped separately, with the former at the top

You can test messages in the same way as outlined in #46, which itself includes a couple of user requests. Assets can be tested in a similar way, using `title`, `description` and `quantity`, e.g.:
```
curl --request POST \
 --url http://localhost:3001/api/assets
 -H 'Content-Type: application/json'
 --data '{ 
   "title": "test 1",
   "description": "desc 1",
   "quantity": 1
   }'
```


jd2rogers2 (approved):
thanks!

PR #86, opened 2021-12-04
add eslint to client and fix react test

jd2rogers2:
## Dev Summary
- re-configuring eslint to be helpful
  - was previously giving us errors in all of the `build` files which made the output too hard to read
- fixes a bunch of missing/extra semis
- adding a line to the pull request template to enforce running lint and test commands
- fixes our react app.test to check for our logo rather than text that isn't there (was breaking before)

## Test Plan

1. `cd client && npm run checks`
see that there are no eslint errors or jest errors
2. remove a `;` from any component (i.e. `client/src/AboutUs.tsx`) and run `npm run checks` again
see that we get an eslint error and therefore eslint is working

## Resources

- https://www.youtube.com/watch?v=bfyI9yl3qfE
- https://eslint.org/docs/rules/


client/src/Header.tsx
```diff
@@ -73,7 +73,6 @@ function Header() {
                     </NavLink>
                     <NavLink className={classes.navLink} to="/about_us">About Us</NavLink>
                     <NavLink className={classes.navLink} to="/how_it_works">How It Works</NavLink>
-                    <NavLink className={classes.navLink} to="/library_and_forum">Library & Forum</NavLink>
```
jd2rogers2, line 76: we don't have this page any more, not sure how this piece stuck around

server/src/assets/assets.controller.ts
```diff
@@ -14,8 +14,8 @@ import {
 import { AssetsService } from './assets.service';
 import { CreateAssetDto } from './dto/create-asset.dto';
 import { UpdateAssetDto } from './dto/update-asset.dto';
+// authGaurd excluded because it is outside of the scope of this PR
 // import { JwtAuthGuard } from '../auth/guards/jwt-auth.guard';
-//authGaurd excluded because it is outside of the scope of this PR
```
jd2rogers2, line 18: comments go above the line they're referring to

.github/pull_request_template.md
```diff
@@ -26,6 +26,13 @@ You should include screenshots to help the reviewer understand what to look for
 changes are includes. GIFs help immensely when explaining interactions and animations.
 -->
 
+preReqs:
+
+[ ] following command succeeds `cd client && npm run checks && cd ../server && npm run checks`
```
jd2rogers2, line 31: server side scripts, eslint config, etc. will come in a follow up PR
but doing here so i don't forget to add later

client/src/types/material-ui.d.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
yndajas, line 1: Think you can delete this? Both imports are used

yndajas (changes requested):
- One line that can probably be removed
- Might be obvious, but `npm run checks` failed until I ran `npm start` - not sure if the client needs to be running or if I just need to have run it at least once, but might be good to include that in the testing details

client/src/types/material-ui.d.ts
jd2rogers2: they're actually not, we are creating our own types with the same names
(which i'm not sure if this is a good thing, but a bunch of other things break if we remove this module declaration or if we remove these imports 🤷, fixing either of those will have to come in another PR)

if you remove my comment and run `npm run lint` from client folder, you should see warnings. lmk if not

jd2rogers2:
> Might be obvious, but npm run checks failed until I ran npm start - not sure if the client needs to be running or if I just need to have run it at least once, but might be good to include that in the testing details

hmmm, no idea why that's the case. i don't think that should happen
in fact, i didn't run `start` until after all of my changes as an after thought

yndajas:
> > Might be obvious, but npm run checks failed until I ran npm start - not sure if the client needs to be running or if I just need to have run it at least once, but might be good to include that in the testing details
> 
> hmmm, no idea why that's the case. i don't think that should happen in fact, i didn't run `start` until after all of my changes as an after thought

Strange - it's working now without the client running, but this is after having run the client at least once. It said something about missing a checks folder I think, which doesn't quite make sense to me

client/src/types/material-ui.d.ts
yndajas: Yep, that gave warnings. Interesting!

yndajas (approved):

PR #87, opened 2021-12-04
add eslint to server and prettierify server

jd2rogers2:
## Dev Summary

this PR is being compared to branch from https://github.com/Nonprofit-Exchange-Hub/web-app/pull/86
this PR should wait until after that PR is reviewed/landed

- moved old eslint file (removed in https://github.com/Nonprofit-Exchange-Hub/web-app/pull/86) into server
- ran `npm run format`
  - which changed predominantly whitespace throughout server

## Test Plan

1. `cd server && npm run checks`
2. change auth.controller to have...
```
  constructor(private authService: AuthService) {
    
  }
```
3. run `npm run checks`
see that we get an error and that eslint is therefore working

## Resources

- https://www.youtube.com/watch?v=bfyI9yl3qfE
- https://eslint.org/docs/rules/


server/package.json
```diff
@@ -14,12 +14,13 @@
         "start:debug": "nest start --debug --watch",
         "start:dev:db": "sudo docker-compose up",
         "start:prod": "node dist/main",
-        "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
+        "lint": "eslint src/**/*.ts",
```
jd2rogers2, line 17: simplifying and removing `--fix` (we have a `format` script for that)

yndajas (approved):
All looks good

yndajas (approved):
Looks good, although it's slightly confusing that the tests pass, then it pauses for a bit, then it says there's an error

Screenshot 2021-12-04 at 22 26 50

jd2rogers2:
> Looks good, although it's slightly confusing that the tests pass, then it pauses for a bit, then it says there's an error
> 
> Screenshot 2021-12-04 at 22 26 50

that's expected, though i do see what you mean.
combining to one script is mostly just convenience (will be good down the road when we want failed tests to block deploys)
`npm run checks` runs 2 commands. first the `test` command, then `lint`. so when we make a change that should cause a lint error. the output from the first command passes but then the second fails

@yndajas would it be clearer if we make the PR template run each command separately maybe?

yndajas:
Ahh of course! Might be clearer but it is nice just having a single `checks` command. What if there was an `echo` in the middle along the lines of `\nTests complete\nNow running lint checks\n`?

PR #91, opened 2021-12-06
persist user

jd2rogers2:
## Dev Summary

we want logged in user to persist across page refreshes
here's the end to end workflow and some details around the whys:

1. in `client/src/Login.tsx`: user logs in same as before. but now we send along a header with fetch `credentials: 'include',` which makes a handshake with the BE and tells the browser that any cookies we get back with the response are okie doke. in the future it'll be good to set this to `credentials: 'same-origin',` but localhost:3000 and localhost:3001 in development
2. in `server/src/auth/auth.controller.ts` `login()`: on BE reception of request create a jwt with user info, and set into (signed) cookie. add the cookie to the response, and also respond with user object. during research for this ticket i learned the JWTs are encoded and can't be altered (without us knowing about it) but the data is not encrypted (so client could see what's in the body). with this info we no longer need to encode the user data that we respond with and that we'll store in react state. but we do still get a win out of encoding the data we put in the cookie (which is also encrypted) because now we'll know that the cookie+jwt that we receive are authentic and were created by us on the BE
3. so when browser gets back this response it will automatically set the jwt user info into a cookie. and also because we received the user data we set that into state
4. in `client/src/providers/UserProvider.tsx` `useEffect`: on refresh, as our component tree loads, the useEffect in the UserProvider will hit and send a request to `/auth/session`. the browser will automatically send the cookie with this request. 
5. before `server/src/auth/auth.controller.ts` `session()` hits, the `CookieAuthGuard` decorator hits and checks if the cookie is there and if it's valid (not expired, signed by us) and if so, adds the user to the request data (all this in `server/src/auth/guards/cookie-auth.guard.ts` `canActivate`)
6. back in `server/src/auth/auth.controller.ts` `session()`: this responds with the user so we can have that in state on page load
7. `UserProvider` sets user into state so that we are already logged in on page load
8. in `setTimeout` in `UserProvider`: if the jwt is about to expire we make another request to `/auth/session` to get a new one
9. in `client/src/Header.tsx`: new logout button that sends request to `/auth/logout`. also removes user from state
10. in `server/src/auth/auth.controller.ts` `logout`: clears cookie aka (i think) sends new cookie with value of null back to browser

## Test Plan

preReqs (to be checked by reviewer):

- [ ] following command succeeds:
`cd client && npm run checks && cd ../server && npm run checks && cd ..`
expect tests to pass
- [ ] ensure all files have been formatted:
`cd client && npm run format && cd ../server && npm run format && cd ..`
expect `git status` to show 0 files changed

repro steps:

1. login
open dev tools > network > login request
expect to see response headers with cookie
switch to application tab
expect to see cookie listed under localhost:3000
2. refresh page
expect to still be logged in
3. change setTimeout time to be `5 * 1000`
open dev tools > network tab
expect to see requests to /auth/session every 5 seconds and login state persists
4. party

## Resources

- https://github.com/nestjs/jwt
- https://stackoverflow.com/questions/15995919/how-to-use-curl-to-send-cookies
- https://docs.nestjs.com/security/authentication
- https://docs.nestjs.com/techniques/cookies
- https://blog.logrocket.com/jwt-authentication-best-practices/#brief-introduction-to-jwt
- https://scotch.io/bar-talk/why-jwts-suck-as-session-tokens
- https://stackoverflow.com/questions/33112299/how-to-delete-cookie-on-logout-in-express-passport-js
- https://github.com/expressjs/cookie-parser
- https://docs.nestjs.com/controllers#library-specific-approach
- https://stackoverflow.com/questions/36824106/express-doesnt-set-a-cookie
- https://stackoverflow.com/questions/7042340/error-cant-set-headers-after-they-are-sent-to-the-client
- https://expressjs.com/en/api.html#res.send
- https://medium.com/swlh/7-keys-to-the-mystery-of-a-missing-cookie-fdf22b012f09
- https://docs.nestjs.com/security/cors
- https://github.com/expressjs/cors#configuration-options
- https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials
- https://stackoverflow.com/questions/52285591/signing-cookies-in-express
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict


server/src/auth/guards/cookie-auth.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+    try {
+      user = await this.jwtService.verify(jwt, { secret: jwtConstants.secret });
+    } catch (_e) {
+      return false;
+    }
+
+    request.user = user;
+    return true;
+  }
+}
```
jd2rogers2, line 35: would've loved to been able to get `super.canActivate(context);` to work but just couldn't 😢

client/src/providers/UserProvider.tsx
```diff
@@ -2,14 +2,50 @@ import * as React from 'react';
...
+
+  function setUserTimeout(isInitialCall: boolean): void {
+    setTimeout(async () => {
+      if (!isInitialCall) {
+        const newUser = await fetchUser();
+        setUser(newUser);
+      }
+      setUserTimeout(false);
+    }, 59 * 60 * 1000);
+  }
```
jd2rogers2, line 36: the value of the user object wouldn't actually change, so we wouldn't see any rerenders down the line
but idk if we need to be `setUser`ing...

server/src/main.ts
```diff
@@ -1,19 +1,28 @@
...
+
 async function bootstrap() {
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
   app.setGlobalPrefix('api');
-  app.enableCors();
+  app.enableCors({
+    credentials: true,
+    // TODO get env related base url
+    origin: 'http://localhost:3000',
+  });
```
jd2rogers2, line 18: allows requests from clients only with origins of localhost
https://docs.nestjs.com/security/cors
https://github.com/expressjs/cors#configuration-options

server/src/auth/auth.controller.ts
```diff
@@ -1,26 +1,42 @@
...
   constructor(private authService: AuthService) {}
 
-  @Post('test')
+  @Post('login')
   @UseGuards(LoginAuthGuard)
-  async authTest() {
-    return {
-      message: 'ヽ(•‿•)ノ',
-    };
+  async login(@Request() request, @Response({ passthrough: true }) response): Promise<void> {
```
esteban-gs, line 16: We'll probably want to use the types for both `request` and `response`: https://docs.nestjs.com/controllers#request-object

server/src/auth/auth.controller.ts
```diff
@@ -1,26 +1,42 @@
...
+    response
+      .cookie(COOKIE_KEY, jwt, {
+        domain: 'localhost',
+        expires: new Date(new Date().getTime() + 60 * 60 * 1000), // 1 hour
+        httpOnly: true,
+        path: '/',
+        secure: true,
+        signed: true,
+      })
+      .send({ user });
```
esteban-gs, line 28: we probably need to set the `samesite` attribute to `strict` to avoid having the auth cookie sent along with 3rd party requests: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict.

jd2rogers2:
@esteban-gs addressed your comments and updated UserProvider, ready for another look if you want

client/src/providers/UserProvider.tsx
```diff
@@ -2,14 +2,59 @@ import * as React from 'react';
...
+  user: User | null,
+  shouldFetch?: boolean,
+  shouldStartTimer?: boolean,
+) => Promise<void>;
+type UserContextT = [User | null, SetUser];
 
-export const UserContext = React.createContext<UserContextT>([null, (user: any) => {}]);
+export const UserContext = React.createContext<UserContextT>([
+  null,
+  async function (user: User | null, shouldFetch?: boolean, shouldStartTimer?: boolean) {},
```
esteban-gs, line 14: I'm not as versed with React ways as I am with Angular. Why is the function body here empty?

jd2rogers2: yeah it's an odd pattern
this is basically the initial value of the context. we don't ever use it though, what's used is what's passed to the provider below 
`return <UserContext.Provider value={[user, setUserTimeout]}>{children}</UserContext.Provider>;`
so we basically overwrite it on the first render with the `setUserTimeout` timeout function

https://reactjs.org/docs/context.html

esteban-gs (approved):

PR #92, opened 2021-12-06
Create transactions resource

DXRovang:
## Dev Summary
adding the transactions resource here. this is a key part of our data model. this is how users will claim/give assets. 
this includes the standard parts of a new resource (controller, service, module, entity, dtos, etc.)

## Test Plan

repro steps:

1. `npm run start:dev`
2. seed some data if you need
```
curl -X POST localhost:3001/api/users \
-H "Content-Type: application/json" \
-d '{ "first_name": "test", "last_name": "user", "password": "password", "email": "test@test.com" }'

curl -X POST localhost:3001/api/users \
-H "Content-Type: application/json" \
-d '{ "first_name": "test", "last_name": "user2", "password": "password", "email": "test2@test.com" }'

# except below should have the id that one of the user create requests returned
curl -X POST localhost:3001/api/assets \
-H "Content-Type: application/json" \
-d '{ "title": "thing 1", "description": "its a thing", "type": "donation", "quantity": "1",
"poster": { "first_name":"test", "last_name":"user", "email":"test@test.com", "id": "18" } }'

curl -X POST localhost:3001/api/organizations \
-H "Content-Type: application/json" -d '{
  "name": "name",
  "description": "description",
  "website": "website.com",
  "address": "address",
  "phone": "phone",
  "city": "seattle",
  "state": "wa",
  "ein": "1",
  "tax_exempt_id": "1"
}'
```

3. create transaction
```
# below should use the other created user's id
jamesrogers@jamesrogpremiseMBP-2 web-app % curl -X POST localhost:3001/api/transactions \
-H "Content-Type: application/json" -d '{ "donater_user": "19", "asset": "1" }'
{"donater_user":"19","asset":"1","id":3,"status":"IN_PROGRESS","created_date":"2022-01-22"}%      
jamesrogers@jamesrogpremiseMBP-2 web-app % 
```

4. update transaction status and with recipient org
```
jamesrogers@jamesrogpremiseMBP-2 web-app % curl -X PATCH localhost:3001/api/transactions/3 \
-H "Content-Type: application/json" -d '{ "status": "COMPLETED", "recipient": "1"}' 
{"id":3,"status":"COMPLETED","created_date":"2022-01-22","asset":{"id":1,"title":"thing 1","description":"its a thing","type":"donation","condition":"","quantity":1},"recipient":"1"}%             
jamesrogers@jamesrogpremiseMBP-2 web-app % 
```

5. get all transactions
expect there to be one
```
jamesrogers@jamesrogpremiseMBP-2 web-app % curl localhost:3001/api/transactions 
[{"id":3,"status":"COMPLETED","created_date":"2022-01-22","asset":{"id":1,"title":"thing 1","description":"its a thing","type":"donation","condition":"","
```

6. delete the transaction
then get all transactions again and expect it to be empty
```
jamesrogers@jamesrogpremiseMBP-2 web-app % curl -X DELETE localhost:3001/api/transactions/3
jamesrogers@jamesrogpremiseMBP-2 web-app % curl localhost:3001/api/transactions            
[]%       
```

## Resources

- https://typeorm.io/#/eager-and-lazy-relations
- https://docs.nestjs.com/techniques/database#custom-repository
- https://typeorm.io/#/custom-repository
- https://docs.nestjs.com/techniques/validation
- https://github.com/typestack/class-validator


server/package.json
```diff
@@ -33,19 +33,21 @@
         "@nestjs/serve-static": "^2.2.2",
         "@nestjs/typeorm": "^7.1.5",
         "bcrypt": "^5.0.1",
+        "class-transformer": "^0.4.0",
```
jd2rogers2, line 36: i don't see this one being used anywhere
let's remove until it is being used

server/src/transactions/dto/get-transactions-filter.dto.ts
```diff
@@ -0,0 +1,32 @@
...
+import { User } from 'src/users/entities/user.entity';
+import { Organization } from 'src/organizations/entities/organization.entity';
+import { Asset } from 'src/assets/entities/asset.entity';
+
+export class GetTransactionsFilterDto {
+  @IsOptional()
+  @IsString()
+  status: TransactionStatus;
+
+  // unclear as to the value/conflict of having both @IsOptional and ?
```
jd2rogers2, line 12: could be interesting to see if we can figure this out
if you don't find anything online you could `npm run build` then look at the corresponding file in `server/dist/...` to see what it compiles to, make the change and rebuild and see if there is a difference in the file

server/src/assets/entities/asset.entity.ts
```diff
@@ -32,5 +33,10 @@ export class Asset {
   quantity: number;
 
   @ManyToOne(() => User, (user) => user.assets)
+  @JoinColumn()
   poster: User;
+
+  @OneToMany(() => Transaction, (transaction) => transaction.asset, { eager: true })
```
jd2rogers2, line 39: neat, but i don't think we want to get all the asset's transactions every time we get the asset
what about asset display page? we don't need all this extra data there
(that is what `eager: true` does right?)

server/src/assets/entities/asset.entity.ts
```diff
@@ -32,5 +33,10 @@ export class Asset {
   quantity: number;
 
   @ManyToOne(() => User, (user) => user.assets)
+  @JoinColumn()
```
jd2rogers2, line 36: 👍 
good catch, can you do a quick audit of the rest of our entities and see if there are any more of these we missed? put anything in a separate PR
if you already had plans to do something else lmk and i'll throw in an issue

jd2rogers2: https://typeorm.io/#/undefined/creating-a-one-to-one-relation
(for resources list in pr/issue)

server/src/main.ts
```diff
@@ -4,12 +4,13 @@ import { NestFactory } from '@nestjs/core';
...
 import { resolve } from 'path';
 import { AppModule } from './app.module';
+import { ValidationPipe } from '@nestjs/common';
 
 async function bootstrap() {
     const app = await NestFactory.create<NestExpressApplication>(AppModule);
     app.setGlobalPrefix('api');
     app.enableCors();
-
+    app.useGlobalPipes(new ValidationPipe())
```
jd2rogers2, line 13: so this was needed in order to use the `IsNotEmpty` and other decorators?

server/src/organizations/entities/organization.entity.ts
```diff
@@ -31,4 +32,8 @@ export class Organization {
 
   @Column('int')
   tax_exempt_id: number;
+
+  @OneToMany(() => Transaction, (transaction)=> transaction.donater_organization || transaction.recipient, { eager: true})
```
jd2rogers2, line 36: same here, don't think we want the eager here

server/src/transactions/transactions.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  async getTransactionById(id: number): Promise<Transaction> {
+    // NOTE:  findOne is a built in function
+    const found = await this.transactionsRepository.findOne(id);
+    if (!found) {
+      throw new NotFoundException();
+    }
+    return found;
+  }
+
+  async updateTransaction(
```
jd2rogers2, line 36: this should either be `updateTransactionStatus` or be able to update any property
i'm leaning towards the latter unless you see a benefit in having multiple endpoints for updating different properties

server/src/users/entities/user.entity.ts
```diff
@@ -21,7 +22,10 @@ export class User {
 
   @OneToMany(() => Asset, (asset) => asset.poster)
   assets: Asset[];
-  
+
+  @OneToMany(() => Transaction, (transaction)=> transaction.donater_user, { eager: true})
```
jd2rogers2, line 26: remove eager

jd2rogers2: also i'm seeing some missing whitespace 😲 
make sure you are up to date with master (`git pull origin main`, resolve conflicts, (if conflicts) `git add . && git merge --continue`)
and then run `npm run format` 
and might as well run `npm run checks` for good measure too

server/src/transactions/entities/transaction.entity.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @ManyToOne(
+    (_type) => Organization,
+    (organization) => organization.transactions,
+    { eager: false },
+  )
+  @JoinColumn()
+  donater_organization?: Organization;
+
+  @ManyToOne((_type) => Asset, (asset) => asset.transactions, { eager: false })
```
jd2rogers2, line 41: actually this time we might want `eager: true` iiuc

server/src/transactions/entities/transaction.entity.ts
```diff
@@ -0,0 +1,49 @@
...
+  asset: Asset;
```
jd2rogers2, line 42: i think the `JoinColumn` should be here no?
asset_id is in the transactions table, so this table owns the relationship

jd2rogers2: so remove corresponding from asset entity

server/src/auth/get-user.decorator.ts
```diff
@@ -0,0 +1,7 @@
+import { createParamDecorator, ExecutionContext } from "@nestjs/common";
+import { User } from "src/users/entities/user.entity";
+
+export const GetUser = createParamDecorator((_data, ctx: ExecutionContext): User => {
+  const req = ctx.switchToHttp().getRequest()
+  return req.user
+})
```
jd2rogers2, line 7: nice! much more nesty than `request.user = user;` in https://github.com/Nonprofit-Exchange-Hub/web-app/pull/91/files

server/src/assets/entities/asset.entity.ts
```diff
@@ -32,5 +33,10 @@ export class Asset {
...
   poster: User;
+
+  @OneToMany(() => Transaction, (transaction) => transaction.asset, { eager: true })
+  @JoinColumn()
+  transactions: Transaction[];
```
jd2rogers2, line 41: so i think remove eager and joinColumn here

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TransactionStatus } from '../transaction-status.enum';
+import { User } from 'src/users/entities/user.entity';
+import { Organization } from 'src/organizations/entities/organization.entity';
+import { Asset } from 'src/assets/entities/asset.entity';
+
+export class CreateTransactionDto {
+  @IsNotEmpty()
+  donater_user: User;
+
+  donater_organization: Organization;
```
jd2rogers2, line 11: optional here i think no?

server/src/transactions/transactions.service.ts
```diff
@@ -0,0 +1,53 @@
...
+export class TransactionsService {
+  constructor(
+    @InjectRepository(Transaction)
+    private transactionsRepository: Repository<Transaction>,
+  ) {}
+
+  async createTransaction(
+    createTransactionDto: CreateTransactionDto,
+  ): Promise<Transaction> {
+    return await this.transactionsRepository.save(createTransactionDto);
```
jd2rogers2, line 19: don't need the `await` here
`return` has that built in

server/src/transactions/transactions.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    id: number,
+    status: TransactionStatus,
+  ): Promise<Transaction> {
+    const transaction = await this.getTransactionById(id);
+    transaction.status = status;
+    await this.transactionsRepository.save(transaction);
+    return transaction;
+  }
+
+  async deleteTransaction(id: number): Promise<void> {
```
jd2rogers2, line 46: i'm actually not sure what this is but let's figure it out and decide on a pattern right now
https://github.com/Nonprofit-Exchange-Hub/web-app/blob/57ab43bb106ee8b11105fcd2cd5a9a42ffd9271c/server/src/assets/assets.service.ts#L30
we do return something from this delete

jd2rogers2 (commented):
lots of questions cuz you're doing some cool stuff here
nice work!

also though, can you add, to your pr summary, some postman screenshots or snippets with curl/httpie request examples so reviewer can test

server/src/transactions/dto/create-transaction.dto.ts
DXRovang: I was talking to Kevin about this, and we thought that since a Transaction had to have a donater, it should probably be a donater_user who could optionally also have a donater_organization id.   If both donater_user and donater_organization were optional we could have a transaction with no donater.   Does that make sense?

server/src/assets/entities/asset.entity.ts
DXRovang: Yeah, I think you're right.    Looking at this:  https://stackoverflow.com/questions/68992271/meaning-of-using-eagertrue-option-in-typeorm    and the last line "eager loading means related data is automatically fetched."    

So just get rid of the eager on both sides, right?   Or did you mean delete the @OneToMany on the Asset side?   Can I have just a one-sided relation?

server/src/main.ts
DXRovang: Yep!

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  @IsNotEmpty()
+  recipient: Organization;
+
+  // custom message for example, not necessary to code
```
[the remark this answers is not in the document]
DXRovang: No, he didn't talk about that (yet).

server/src/assets/entities/asset.entity.ts
DXRovang: Currently none of the entities in the app use @JoinColumn().    I'm putting this here so I remember it:    https://orkhan.gitbook.io/typeorm/docs/relations   I think I need to digest what's meant by a Join Column.   It looks like what they're saying is that you only want it in a table that has a foreign key, is that correct?

server/src/users/entities/user.entity.ts
DXRovang: git merge won't mess anything up on main?

server/src/assets/entities/asset.entity.ts
jd2rogers2: i just meant to get rid of the `eager` piece

server/src/assets/entities/asset.entity.ts
jd2rogers2: yes, i think it's an explicit way to set the foreign key. otherwise it's left up to nest to figure it out (which is probably fine for our use cases, but doesn't hurt to have it)

server/src/transactions/dto/create-transaction.dto.ts
jd2rogers2: yep, i was referring to just the `donater_organization`. sorry for not being clear

server/package.json
jd2rogers2: actually, we may need it 😕 
just tried to start BE server without it and got errors 🤷

DXRovang: Reinstalled it.

server/src/assets/entities/asset.entity.ts
jd2rogers2: https://github.com/Nonprofit-Exchange-Hub/web-app/issues/127

jd2rogers2 (commented):
hmm, now when i start the server i get
```
[Nest] 10411   - 01/08/2022, 1:58:51 PM   [PackageLoader] The "class-transformer" package is missing. Please, make sure to install this library ($ npm install class-transformer) to take advantage of ValidationPipe. +4ms
```
so maybe we do need that dependency

server/src/users/entities/user.entity.ts
```diff
@@ -1,6 +1,7 @@
 import { Entity, Column, PrimaryGeneratedColumn, OneToMany } from 'typeorm';
 import { Asset } from '../../assets/entities/asset.entity';
 import { Message } from '../../messages/entities/message.entity';
+import { Transaction } from '../../transactions/entities/transaction.entity'
```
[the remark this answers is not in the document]
DXRovang: That throws the following error:

src/users/entities/user.entity.ts:26:20 - error TS1361: 'Transaction' cannot be used as a value because it was imported using 'import type'.

26   @OneToMany(() => Transaction, (transaction)=> transaction.donater_user)

server/src/main.ts
```diff
@@ -16,11 +17,10 @@ async function bootstrap() {
     // TODO get env related base url
     origin: 'http://localhost:3000',
   });
-
+  app.useGlobalPipes(new ValidationPipe());
```
esteban-gs, line 20: 👍🏾

server/src/transactions/entities/transaction.entity.ts
```diff
@@ -0,0 +1,40 @@
...
+    enum: TransactionStatus,
+    default: TransactionStatus.IN_PROGRESS,
+  })
+  status: TransactionStatus;
+
+  @Column({
+    type: 'date',
+    default: new Date().toLocaleString(),
+  })
+  created_date: Date;
```
esteban-gs, line 23: ### Should fix:
Here, you are using Javascript dates, but we should let TypeORM use their SQL timestamp decorator:

https://typeorm.io/#/decorator-reference/createdatecolumn

```typescript
@Entity()
export class User {

    @CreateDateColumn()
    createdDate: Date;

}
```

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,29 @@
+import { IsNotEmpty, IsEnum, IsOptional } from 'class-validator';
+import { TransactionStatus } from '../transaction-status.enum';
+
+import type { User } from 'src/users/entities/user.entity';
+import type { Organization } from 'src/organizations/entities/organization.entity';
+import type { Asset } from 'src/assets/entities/asset.entity';
+
+export class CreateTransactionDto {
+  @IsNotEmpty()
+  donater_user: User;
```
esteban-gs, line 10: ### Fix:
Should be:
``` typescript
 @IsNotEmpty()
  donater_user_id: number;
```

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  @IsOptional()
+  donater_organization: Organization;
```
esteban-gs, line 13: ### Fix:
### Fix:
Should be:
``` typescript
 @IsNotEmpty()
  donater_organization_id: number;
```

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  @IsOptional()
+  recipient: Organization;
+
```
esteban-gs, line 17: Same as above, only the foreign key is required

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+  // custom message for example, not necessary to code
+  @IsNotEmpty({ message: 'asset_id is required' })
+  asset: Asset;
```
esteban-gs, line 20: Same as above, only the foreign key is required

server/src/transactions/dto/get-transactions-filter.dto.ts
```diff
@@ -0,0 +1,5 @@
+import { PartialType } from '@nestjs/mapped-types';
+
+import { CreateTransactionDto } from './create-transaction.dto';
+
+export class GetTransactionsDto extends PartialType(CreateTransactionDto) {}
```
esteban-gs, line 5: ### Recommended:
It looks like we're not using this filter DTO at the moment. I recommend dropping it if we're not using it. If you remove it, the `Get()` controller method will return all records in this table.

Alternatively, if you want to implement, you should probably create properties for it instead of extending from `CreateTransactionDto`. [Example](https://jasonwhite.xyz/posts/2020/10/19/crud-api-w/-server-side-pagination-with-nestjs-angular-pagination/)

server/src/transactions/transactions.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    id: number,
+    updateTransactionDto: UpdateTransactionDto,
+  ): Promise<Transaction> {
+    const transaction = await this.getTransactionById(id);
+    const newTransaction = await this.transactionsRepository.save({
+      ...transaction,
+      ...updateTransactionDto,
+    });
+    return newTransaction;
+  }
```
esteban-gs, line 43: ### Recommended:
#### Fix or leave for another PR
It's best practice to only use the `dto` in the transport layer (controller) and pass the entity to the repository. This may require some manual mapping from `dto` to `entity`. There are some libraries that assist with that.

[Stack Overflow on `dto` -> `entity` conversions](https://stackoverflow.com/a/65325058/13862254)
[NestJS auto-mapper library](https://automapperts.netlify.app/docs/introduction/what-why/)

esteban-gs (changes requested):
See comments

server/src/transactions/dto/create-transaction.dto.ts
jd2rogers2: hmmm, are you sure?
would nest know to add a `.donater` attr or function to the transation model then? now that i think about it, idk how to control the model in nest, like if i wanted to write a custom function like `asset.close()` (that updates the status to all the transactions related to that asset, how would i do that?

server/src/transactions/dto/get-transactions-filter.dto.ts
jd2rogers2: but the properties would be the same as the `CreateTransactionDto` no? (except optional, hence [PartialType](https://docs.nestjs.com/techniques/validation#mapped-types))

and i forgot to write a test case for it but we could use the filterDTO, i.e.
```
jamesrogers@jamesrogpremiseMBP-2 web-app % curl -X GET "localhost:3001/api/transactions"     [{"id":5,"status":"IN_PROGRESS","created_date":"2022-01-29T19:11:38.779Z","asset":{"id":3,"title":"thing 2","description":"its also thing","type":"donation","condition":"","quantity":1}},{"id":4,"status":"COMPLETED","created_date":"2022-01-26T01:08:55.808Z","asset":{"id":1,"title":"thing 1","description":"its a thing","type":"donation","condition":"","quantity":1}}]%       
jamesrogers@jamesrogpremiseMBP-2 web-app % curl -X GET "localhost:3001/api/transactions?status=IN_PROGRESS" \
-H "Content-Type: application/json"
[{"id":5,"status":"IN_PROGRESS","created_date":"2022-01-29T19:11:38.779Z","asset":{"id":3,"title":"thing 2","description":"its also thing","type":"donation","condition":"","quantity":1}}]%
jamesrogers@jamesrogpremiseMBP-2 web-app % 
```

server/src/transactions/transactions.service.ts
jd2rogers2: ok, i'm listening
sounds interesting
would have to do more research though, and also definitely a separate PR since we'd need to clean this up in a bunnch of other spots as well
https://github.com/Nonprofit-Exchange-Hub/web-app/issues/136

server/src/transactions/dto/get-transactions-filter.dto.ts
esteban-gs: I actually, just noticed that there is some filtering happening in the service:
https://github.com/Nonprofit-Exchange-Hub/web-app/blob/fe39084b837da3e368914d190ef622c55e7f35c2/server/src/transactions/transactions.service.ts#L22


But in any case, it's best practice to separate the filtering logic (the createDto is playing that role) from the createDto model, otherwise, you will find yourself making the createDto include more props than the client needs to pass for creating and vice-versa. 

Your call, at this point, it's probably not that important.

jd2rogers2: yeah i just added that 😅  wasn't there before
but that's what i had intended, didn't realize it wasn't working before (had assumed it was since we're doing this in other places)

do you mean separated in some other way? to me these already seem separated the way you want createDto for creating and this getDto has optional properties that we can filter by (and add to like if we wanted to paginate and add a limit property or something)

server/src/transactions/dto/create-transaction.dto.ts
esteban-gs: Looking at your example, 

This:
```bash
curl -X POST localhost:3001/api/assets \
-H "Content-Type: application/json" \
-d '{ "title": "thing 1", "description": "its a thing", "type": "donation", "quantity": "1",
"poster": { "first_name":"test", "last_name":"user", "email":"test@test.com", "id": "18" } }'
```

can be replaced by example blow, if you change the createDto to use poster_id instead of a whole new poster object. The return type will be whatever the repository returns (`transactionEntity`):
```bash
curl -X POST localhost:3001/api/assets \
-H "Content-Type: application/json" \
-d '{ "title": "thing 1", "description": "its a thing", "type": "donation", "quantity": "1",
"poster_id": 18 }'
```

server/src/transactions/dto/create-transaction.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  @IsOptional()
+  @IsEnum(TransactionStatus)
+  status: TransactionStatus;
+
+  @IsOptional()
+  @IsNotEmpty()
+  created_date: Date;
```
esteban-gs, line 28: Don't these cancel each other?

server/src/transactions/dto/create-transaction.dto.ts
esteban-gs: Then, you would have to implement something like: https://github.com/Nonprofit-Exchange-Hub/web-app/compare/create-transactions-resource...sample/create-transactions-resource

Also, if you are going to stick with the `getTransactionsDto` as the get all filtering model, there is some funky formatting we have to do in the front end to send the nested object in the url query params. See [this](https://nickdrane.com/build-your-own-nested-query-string-encoder/)

esteban-gs (approved):

server/src/transactions/dto/create-transaction.dto.ts
jd2rogers2: totally 😆 
removed altogether because the default value should be used and so we don't need to pass anything in

PR #99, opened 2021-12-14
update check/format commands to return to root

yndajas:
Small edit to the check and format commands to return the user to the root `web-app` directory. Without this, you need to remember to run `cd ..` in between the two (chains of) commands


jd2rogers2 (approved):
good call

PR #120, opened 2021-12-27
Validate organizations

zoeblairfriedman:
## Dev Summary

This PR:
**1. Creates an endpoint for the EIN check in the Organizations Controller** (GET "/organizations/ein/:ein"). 
This endpoint will return a ProPublica organization object or `null`. 

An abbreviated valid EIN return example: 
```
{
    "id": 142007220,
    "ein": 142007220,
    "name": "PRO PUBLICA INC",
    "careofname": null,
    "address": "155 AVE AMERICA 13 FL",
    "city": "NEW YORK",
    "state": "NY",
    "zipcode": "10013-0000",
    "exemption_number": 0,
    "subsection_code": 3,
    "affiliation_code": 3,
    "classification_codes": "1000",
    "ruling_date": "2008-02-01",
    "deductibility_code": 1,
    "foundation_code": 15,
 ...
}
```

This will allow a request to be made from the front end to verify the organization prior to a user being allowed to save the instance. 

The object will give the frontend access not only to the EIN and name, but perhaps other helpful information for the future. 

**2. This PR also adds the EINCheck function in the Organizations Service** for the new endpoint to utilize. This makes a fetch request, checking ProPublica's API for that EIN number, returning that organization as saved in their API. 

Originally, I attempted to make the API call using the recommended Axios module per the [Nest documentation](https://docs.nestjs.com/techniques/http-module). James and I spent a lot of time barking up this tree, trying to utilize Observables and struggling with the asynchronous calls. Ultimately, he recommended I use `fetch` instead, which turned out to be a wildly simpler implementation. This way I was able to return an object wrapped in an easy to understand Promise, rather than a more confusing, Observable. 

Furthermore, I originally incorporated the name and EIN cross-checking logic in the service, upon the organization's creation. However, upon review and discussion with James regarding the user experience, we decided to save whatever was sent from the front end in the ```create``` function, and to more simply return the org object from the ```EINCheck``` function. 

## Test Plan
To test this PR, make get requests to the EIN endpoint with both valid, and invalid EINs. 

preReqs (to be checked by reviewer):

- [ ] following command succeeds:
`cd client && npm run checks && cd ../server && npm run checks && cd ..`
expect tests to pass
- [ ] ensure all files have been formatted:
`cd client && npm run format && cd ../server && npm run format && cd ..`
expect `git status` to show 0 files changed

repro steps:

1. Make a GET request via Postman or other http service to: `http://localhost:3001/api/organizations/ein/142007220` With this valid EIN, you should receive an object with many organization fields: 

PPObj

2. Make a GET request with an invalid EIN. A la: `http://localhost:3001/api/organizations/ein/242007220` With this invalid EIN, you should receive nothing: 

null

## Resources
- [Propublica API](https://projects.propublica.org/nonprofits/api/)
- [utilizing exceptions](https://docs.nestjs.com/exception-filters)
- [GET request with params](https://stackoverflow.com/questions/54958244/how-to-use-query-parameters-in-nest-js)
- [making fetch requests in node](https://blog.logrocket.com/5-ways-to-make-http-requests-in-node-js/)


zoeblairfriedman:
Still trucking along with this! A few questions: 

1. I added an EINCheck interface for my checkEIN function. Was that the right way to go? I'm still so new to TS and want to make sure I'm doing this right (or if it was even necessary).

2. Nest has an equivalent Expectation_Failed object I can get from the HTTPStatus helper via [this resource](https://docs.nestjs.com/exception-filters#custom-exceptions), but the status code is 409 and not 417. Is that okay? Or was there a specific reason you were looking for the 417 status code you [referenced here](https://hapi.dev/module/boom/api?v=9.1.4#boomexpectationfailedmessage-data). 

Here's what we're receiving now: 
Successful checkEIN:

SuccessfulyVerified

EIN exists, but name is incorrect:

InvalidName

Organization with this EIN does not exist:

OrgNotFound

3. I'm worried I'm way off on how we'll send back the suggested name pulled from the ProPublica API to the front end. Right now I have it set up to be pulled from the HttpException. Is that a normal practice or am I way off base?

server/src/organizations/EINCheck.ts
```diff
@@ -0,0 +1,5 @@
+export class EINCheck {
+  verified: boolean;
+  einExists: boolean;
+  actualName: string;
+}
```
jd2rogers2, line 5: this is only being used as a type so let's just make it a type
`type EINCheck = {...`
also, it's only used in the org service, so i'd say we can just keep it local there

server/src/organizations/organizations.controller.ts
```diff
@@ -43,11 +43,8 @@ export class OrganizationsController {
   }
 
   @Patch(':id')
-  async update(
-    @Param('id') id: string,
-    @Body() updateOrganizationDto: UpdateOrganizationDto,
-  ): Promise<Organization> {
-    return this.organizationsService.update(+id, updateOrganizationDto);
+  update(@Param('id') id: string, @Body() body: UpdateOrganizationDto): Promise<Organization> {
+    return this.organizationsService.update(+id, body);
```
jd2rogers2, line 47: if your only changes in this file are a var name change then maybe it's better to just not touch it
also, i'd argue against this var name change as we seem to have this convention in the rest of your methods

jd2rogers2: `git checkout origin/main -- server/src/organizations/organizations.controller.ts` will put it right back for ya

server/src/organizations/organizations.module.ts
```diff
@@ -5,7 +5,8 @@ import { OrganizationsController } from './organizations.controller';
 import { Organization } from './entities/organization.entity';
 
 @Module({
-  imports: [TypeOrmModule.forFeature([Organization])],
+  imports: [
+    TypeOrmModule.forFeature([Organization])],
```
jd2rogers2, line 9: also don't need this in your diff, not a change relevant to the scope of this PR

jd2rogers2 (commented):
> Still trucking along with this! A few questions:
> 
> 1. I added an [EINCheck interface](https://github.com/Nonprofit-Exchange-Hub/web-app/blob/validate-organizations/server/src/organizations/EINCheck.ts) for my checkEIN function. Was that the right way to go? I'm still so new to TS and want to make sure I'm doing this right (or if it was even necessary).
> 2. Nest has an equivalent Expectation_Failed object I can get from the HTTPStatus helper via [this resource](https://docs.nestjs.com/exception-filters#custom-exceptions), but the status code is 409 and not 417. Is that okay? Or was there a specific reason you were looking for the 417 status code you [referenced here](https://hapi.dev/module/boom/api?v=9.1.4#boomexpectationfailedmessage-data). Strangely it _should_ be returning a 417...[Expectation Failed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417)
> 3. I'm worried I'm way off on how we'll send back the suggested name pulled from the ProPublica API to the front end. Right now I have it set up to be pulled from the HttpException. Is that a normal practice or am I way off base?

1. a `type` or `interface` would be fine, yep. not `class` though. that implies we're using it for OO and not just typescript
2 & 3. ok, good insight here, thanks for trying to work through that. maybe i/we came to the wrong conclusion last time we talked. here's what i'm thinking for our workflow
- user is filling out organization sign up form
- the submit button is disabled until we make a call with the EIN that responds with the propublica org obj
- if the names don't match we do a `did you mean?` modal
- now they can submit

so this will mean we need 2 separate endpoints. one for the check and another for the create. how's this sound to you?

server/src/organizations/EINCheck.ts
zoeblairfriedman: Copy that. Feeling a little dumb on this one! I think I'd looked in my notes and forgotten that you can make a type w/o it being a class.

server/src/organizations/organizations.controller.ts
zoeblairfriedman: Will do!

server/src/organizations/organizations.module.ts
zoeblairfriedman: copy!

zoeblairfriedman:
Here is an expanded version of your flow just to make sure I understand: 

#1: User enters their EIN & Org Name
#2: User clicks some sort of "Verify EIN" button next to those inputs that routes to new endpoint
#3a: If correct: modal pops up and says "EIN & Name Verified", Submit button enabled.
#3b: If EIN exist but name differs: modal pops up and says "Did you mean GOODWILL ALLIANCE?"
  #3b-yes: Name input changes to reflect Propublica version, input is now unable to be altered, Submit button enabled
  #3b-no: Modal now says "Please check your EIN number and try again"
#3c: If EIN does not exist: Modal pops up and says "Please check your EIN number and try again"

On that note, Propublica stores all of the org names in full caps. Should the checker logic say: "If the user's input converted to all caps matches, it's fine to save as is" or should I save the Propublica version of names? 

And as for creating the separate endpoint, are you thinking a Get request with a query param? Or is there a better route I haven't considered? 

Thanks again!

jd2rogers2:
> Here is an expanded version of your flow just to make sure I understand:
> 
> 1: User enters their EIN & Org Name 2: User clicks some sort of "Verify EIN" button next to those inputs that routes to new endpoint 3a: If correct: modal pops up and says "EIN & Name Verified", Submit button enabled. 3b: If EIN exist but name differs: modal pops up and says "Did you mean GOODWILL ALLIANCE?" 3b-yes: Name input changes to reflect Propublica version, input is now unable to be altered, Submit button enabled 3b-no: Modal now says "Please check your EIN number and try again" 3c: If EIN does not exist: Modal pops up and says "Please check your EIN number and try again"

your walk through is pretty much right, I would say we don't need a `Verify EIN` button and can trigger automatically using a `useEffect`, but we'll worry about all that when we get to the FE side (https://github.com/Nonprofit-Exchange-Hub/web-app/issues/123)

> On that note, Propublica stores all of the org names in full caps. Should the checker logic say: "If the user's input converted to all caps matches, it's fine to save as is" or should I save the Propublica version of names?

in your create org endpoint, just save whatever is passed from the FE. we'll handle the formatting when we do the overwrite/auto-correct of the name
fyi, still some unanswered questions regarding forcing or allowing user to use any name (DBA name) https://nonprofitexchangehub.slack.com/archives/C01G4GR05CJ/p1641253619074200

> And as for creating the separate endpoint, are you thinking a Get request with a query param? Or is there a better route I haven't considered?

for checking the EIN, yes a get request with query param that returns the propublica org obj to the FE sounds good

zoeblairfriedman:
Alright! This all sounds good. On it!

server/src/organizations/organizations.service.ts
```diff
@@ -1,19 +1,45 @@
...
+      );
+      const org = await res.json();
+      if (org) {
+        einObj.validEIN = true;
+        einObj.proPublicaName = org.organization.name;
+      }
+    } catch (err) {
+      console.log(err.message);
+    }
+    return einObj;
```
jd2rogers2, line 42: let's just return the entire propublica org or null
and let's call this function, and the controller function, getPropublicaOrg or something

server/src/organizations/entities/organization.entity.ts
```diff
@@ -27,8 +27,21 @@ export class Organization {
...
+  @AfterInsert()
+  logInsert() {
+    console.log('inserted Org', this.id);
+  }
+
+  // @BeforeInsert()
+  // checkAPI(){
+  //   console.log("checking the API", this.name, this.ein)
+
+  // }
```
jd2rogers2, line 46: do we need all of the changes in this file? if so what are they doing?

jd2rogers2 (commented):
> ## Note:
> I was not sure how to do this from the template guidelines:
> 
> 3. Review merge commit message and remove unnecessary lines
> 
> This would be helpful as I may have made additional changes that I need to explain and justify.

re: "Review merge commit message"
so when you are going to merge your PR, there will be a text input for you to set the commit message, i usually just set it to the title of the PR. and i also remove all of the commits from your PR's branch that are in a second text input

re: and remove unnecessary lines
remove any logs or whitespace changes that don't really do anything
things like that

server/package.json
```diff
@@ -24,6 +24,7 @@
     },
     "dependencies": {
         "-": "0.0.1",
+        "@hapi/boom": "^9.1.4",
```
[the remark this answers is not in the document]
zoeblairfriedman: ah, forgot about that—thanks!

server/src/organizations/entities/organization.entity.ts
zoeblairfriedman: nope, they were just for my experimentation—forgot to remove! Thanks.

server/src/organizations/organizations.service.ts
zoeblairfriedman: Should I create a type that covers all properties of the propublica org? Or is there a TS hack I don't know about that allows you to return a generic object? Just trying to figure out what the return type will be.

zoeblairfriedman: ...ah just found this: https://www.typescriptlang.org/docs/handbook/2/generics.html

zoeblairfriedman: Nevermind, I don't know how helpful that is. Will play around.

zoeblairfriedman: Okay i'm an idiot—disregard!

zoeblairfriedman:
Okay, I took out the unnecessary code and packages, made the requested adjustment, and moved the description to the top of this draft!

server/src/organizations/organizations.service.ts
```diff
@@ -1,19 +1,35 @@
...
+      );
+      const org = await res.json();
+      if (org) {
+        return org.organization;
+      } else {
+        return null
+      }
+    } catch (err) {
+      console.log(err.message)
+    }
```
jd2rogers2, line 32: i think we should also have a return of some sort in here
not sure what makes the most sense
maybe something like
```
if err === not found error
  return err; // (basically throw 404 (expected response if propublica org doesn't exist)
else
  // idk, maybe we want to throw all the time?
  // remind me what the propublica api responds with again
```

server/src/organizations/organizations.service.ts
```diff
@@ -1,19 +1,35 @@
...
-    private organizationsRepository: Repository<Organization>,
+    @InjectRepository(Organization) private organizationsRepository: Repository<Organization>,
   ) {}
 
   async create(createOrganizationDto: CreateOrganizationDto): Promise<Organization> {
-    return await this.organizationsRepository.save(createOrganizationDto);
+    return this.organizationsRepository.save(createOrganizationDto);
+  }
+
+  async checkEIN(ein: number): Promise<object> {
```
jd2rogers2, line 19: let's make a...
```
type PropublicaOrganization = {
  ein: number,
  name: string,
// etc. (only need to add the properties we're using for now, plz include link to the api docs' return object
}
```
to be used in the return type here, instead of `object`

jd2rogers2: not sure if i mentioned this, but meant to...
not a requirement but this function isn't actually doing any checking anymore, just getting the propublicaOrg, so i'd vote for `getPropublicaOrg` naming rather than `checkEIN`

server/src/organizations/organizations.controller.ts
```diff
@@ -42,6 +42,11 @@ export class OrganizationsController {
     return this.organizationsService.findOne(+id);
   }
 
+  @Get('ein/:ein')
+  checkEIN(@Param('ein') ein: number): Promise<object> {
```
jd2rogers2, line 46: suggestion:
```
  @Get('propublicaOrg/:ein')
  getPropublicaOrg(@Param('ein') ein: number): Promise<PropublicaOrg> {
```

server/src/organizations/organizations.service.ts
jd2rogers2: actually, if it responds with 404 when we also want to respond with that, we may not need the try/catch and might just expose this error as our response too

jd2rogers2: if the best option is unclear we can always fix when we do the FE ticket

jd2rogers2 (approved):
couple notes/suggestions, otherwise lgtm

server/src/organizations/organizations.service.ts
jd2rogers2: we actually do want a response, otherwise our FE would just be waiting in some loading state or something
so it should return a 404 or something and then on FE we can handle as needed

Screen Shot 2022-01-08 at 1 00 15 PM

jd2rogers2: also not `invalid json` error in server console when bad ein is passed in

zoeblairfriedman: Hey, sorry I know it's been a while:

Would this work? 

`    } catch (err) {
      return err;
    }`

The value is then: 
`{
    "message": "invalid json response body at https://projects.propublica.org/nonprofits/api/v2/organizations/142007221.json reason: Unexpected token < in JSON at position 0",
    "type": "invalid-json"
}`

Not sure what this would throw us on the front end, but at least it's something for now?

zoeblairfriedman: Or is the goal to find out how to turn it into a 404?

server/src/organizations/organizations.controller.ts
zoeblairfriedman: good call!

server/src/organizations/organizations.service.ts
zoeblairfriedman: copy that.

server/src/organizations/organizations.service.ts
zoeblairfriedman: cause I could throw a NotFoundException and that would return a 404. Maybe that's the move.

zoeblairfriedman: Okay @jd2rogers2 I went for the NotFoundException route. If that work for you I think it's ready to merge.  I made the other changes if you want to double check!

PR #121, opened 2021-12-28
use subscribe with handlers

jd2rogers2:
see my requests on the right
first fails maybe with bad EIN 🤷 
second is success

on the left see console log output 
first:
is failed response
comes from inside the `error` function we pass to `subscribe`
starts like...
```
org <ref *1> Subscriber {
  closed: false,
```
second:
also from first request
error is `something wrong occurred: Error: Request failed with status code 404`
404 = not found, what we'd expect from a bad EIN
third:
is success, comes from inside `next` function we pass to `subscribe`
starts like...
```
next return {
  status: 200,
  statusText: 'OK',
  headers: {
...
  data: {
    organization: {
      id: 142007220,
      ein: 142007220,
```

Screen Shot 2021-12-27 at 9 57 24 PM


jd2rogers2:
@zoeblairfriedman so some notes on this:
`https://projects.propublica.org/nonprofits/api/v2/search.json?q=propublica` you had this URL hardcoded in your org service. everything after the `?` are "search params" which are pieces of the URL that the api server parses out and uses to filter out the selection of data to return
so having `propublica` hardcoded definitely wasn't helping

here's the api docs which have a bit on that and the response shape of the `/organizations/:ein_id.json` that you'll need
https://projects.propublica.org/nonprofits/api

these httpModule docs were misleading :( https://docs.nestjs.com/techniques/http-module

here's where i got my solution from https://rxjs.dev/guide/observable

but yeah not sure if i like subscribe so going to play with it a bit more

PR #128, opened 2022-01-08
Danny b/add authguard to categories

dbrandt1990:
Added JwtAuthGuards to categories resource. I did not apply auth guards to the controller actions for searching categories. This ensures a user can still find categories but will need to be authorized to create, edit, or delete categories.

Here is a POST, PATCH, and DELETE call in Postman to show the unauthorized 401 code after sending the request:

category patch unauth

category delete unauth

GET all categories and GET by id are still accessible.

categories get with auth

catergory by id

### DIFF
--- a/server/src/categories/categories.controller.ts
+++ b/server/src/categories/categories.controller.ts
@@ -8,12 +8,14 @@ import {
   Delete,
   HttpException,
   HttpStatus,
+  UseGuards
 } from '@nestjs/common';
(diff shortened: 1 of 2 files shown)


jd2rogers2 (changes requested):
you'll need to merge in `main`
i actually just deleted the jwt guard in lieu of a cookie guard
but also, idk if these are the guards we're looking for
we want like internal to NEH admin access for updating these categories, not just any logged in user

Issue #83, opened 2021-11-24
frontend file structure


jd2rogers2:
we have 0 file structure on the frontend
let's fix that
some sort of hybrid approach between modularizing by feature and modularizing by function is probably the right way to go
so could be something like...

-featureA
---displayComponents
---hooks
-----useFeatureAResource
---containers
---index.ts
-pageA
---pageA.tsx
---constants.ts
---index.ts
-featureB
....

or....

-components
---featureA
-----containers
---featureB
-----featureB.tsx
-----index.tsx
...
-hooks
---useAsset
-----useAsset
-----utils.ts
-views
---pageA

let's also remove `CookiePolicy`

PR #133, opened 2022-01-18
frontend file structure and routing

yndajas:
## Dev Summary

A few changes, including fixing #83 
- removes cookie policy at James' request
- tidies up frontend file structure (#83), with most files in `/client/src` moved to `/client/src/features` or `/client/assets/sharedComponents` and their subfolders
- refactors route definition into separate `routes` file (in `/src`) where all component imports are handled, and all paths are defined. Components with hard-coded paths now import this and draw paths from here, meaning only one source needs updating if paths change. There are a few links to components/pages that don't exist yet - I've added comments where relevant noting these should be updated once set up
- renames previous `Router` component `Main`, wraps it in `main` tags, and cleans up the `Route`s within using the `routes` file and clearer ordering
- kebab cases routes
- tidies up imports in frontend components

## Test Plan

The only functional changes are the kebab cased routes, which you can see in the various navigation links. Everything else is functionally the same, and all is working in my own tests, which are simple:
1. boot up backend: `cd server && npm run start:dev && cd ..`
2. boot up frontend: `cd client && npm run start && cd ..`
3. navigate around different pages (at [localhost:3000](localhost:3000)) without encountering errors

## Resources

I found an npm package called [madge](https://github.com/pahen/madge) which has CLI functionality to create a nested list or a tree graph of component dependencies, as below. This was pretty helpful as a starting point for thinking about hierarchies - which files belong nearer the top, which belong underneath others and which are shared across fairly unrelated components.

web-app_dependencies

https://user-images.githubusercontent.com/40244233/149857618-1a91d1ab-cb71-4148-a391-c02d990cd194.mov


client/src/routes.tsx
```diff
@@ -0,0 +1,125 @@
+import Home from './features/home/Home';
```
jd2rogers2, line 1: 😍

jd2rogers2 (approved):
so much better! thank you

Issue #114, opened 2021-12-27
create asset in DB from FE


jd2rogers2:
very similar to https://github.com/Nonprofit-Exchange-Hub/web-app/issues/112
makes POST request after user fills out form on page http://localhost:3000/need/goods/
on create should be taken to single asset display page
we should also be able to see this newly created asset is in the DB

ObieMunoz:
closed with PR #139

PR #139, opened 2022-02-06
Handle need asset form submission

ObieMunoz:
Updates #114 with functional POST request.
Update FE condition "Like New" -> "Like new" to resolve a BE error with the formData


client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -69,6 +70,24 @@ function NeedForm() {
     }));
   };
 
+  const handleSubmit = async (evt: React.FormEvent) => {
+    evt.preventDefault();
```
jd2rogers2, line 74: i don't think you need the preventDefault() and therefore not the evt arg either

client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -69,6 +70,24 @@ function NeedForm() {
...
+    const res = await fetch('http://localhost:3001/api/assets', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(formData),
```
jd2rogers2, line 80: not required but you could
log in
get the current user from `UserContext` similar to how we do in `client/src/features/users/Inbox/Inbox.tsx`
then spread the user.id into the formData as the `poster_id` property

jd2rogers2 (approved):
couple small nits but otherwise looks good!

PR #144, opened 2022-03-01
feat: categories FE/BE hookup

### LINKED ISSUE #115: get categories from BE
very similar to https://github.com/Nonprofit-Exchange-Hub/web-app/issues/112
our create asset form has hardcoded categories in the category dropdown on http://localhost:3000/need/goods/
we should get this list from the BE
same with categories filter list on assets index view (http://localhost:3000/assets)

yndajas:
## Dev Summary
[Mocks](https://www.figma.com/file/KE2Sw6ob07agN25RmeNuFg/GiveCycle-wireframes?node-id=1936%3A3444)

Retrieves categories from the backend and populates the field on the goods request form (closes #115)

## Test Plan

1. Load both servers
2. Create some categories, e.g. in Postman, import from [this link](https://www.getpostman.com/collections/368919bc2613efca14fb) and run the post requests (1-5). This will create four categories with `applies_to_assets` set to true, and one with `applies_to_assets` set to false

Screenshot 2022-03-01 at 03 26 13

3. Navigate to the [Share a Need: Goods form](http://localhost:3000/need/goods)
4. Click on the categories input

**Expected:** four categories show, as below

Screenshot 2022-03-01 at 03 27 52

## Comments

Had a couple of user-related warnings preventing pushing (because of the Husky hooks). Commented out one of these where I don't think it's yet being used and is incomplete anyway (reset password), but for the other (in client/src/providers/UserProvider.tsx) I just disabled ESLint for that line. Had a play around to try to fix it, but couldn't see a suitable solution from some quick research. Most suggest moving definition of the function called in the `useEffect` block inside the block, but it's used in the return statement outside that block


client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -53,10 +65,16 @@ const initialFormData: ShareANeedData = {
   deliveryMethod: '',
 };
 
-function NeedForm() {
+function NeedForm(): JSX.Element {
   const [formData, setFormData] = React.useState<ShareANeedData>(initialFormData);
+  const [categories, setCategories] = React.useState<Option[]>([]);
   const history = useHistory();
 
+  (async function () {
```
jd2rogers2, line 73: wrap this in a useEffect with 0 dependencies
(will only fire once on component mount)
will still need the iife

client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -10,11 +11,22 @@ import {
...
-  { value: 'peaches', text: 'Peaches' },
-  { value: 'pears', text: 'Pears' },
-];
+import type { Option } from '../../../types';
+
+const fetchCategories = async (): Promise<Option[]> => {
+  const res = await fetch('http://localhost:3001/api/categories?applies_to_assets=true');
+  const data = await res.json();
+
+  const categories = await data.map((category: any) => {
```
jd2rogers2, line 20: don't need the `await` here
also let's make a `Category` type (can basically be a copy of the BE entity) to use here instead of `any`

client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -10,11 +11,22 @@ import {
...
+    const value = category.name.toLowerCase();
+    const text = category.name;
+
+    return { value, text };
```
jd2rogers2, line 24: i _think_ if we add a `id?: string` to `Option` type and put the id in this object, it'll get passed to BE on form submit and then automatically set the asset<>category relationship

client/src/features/action/needs/NeedFormGoods.tsx
jd2rogers2: (otherwise i think this would run on every rerender)

jd2rogers2 (approved):
couple smallish requests but otherwise looks good!!

PR #147, opened 2022-03-09
got assets from BE to FE

bidiaz101:
## Dev Summary

[Mocks](link)

Set up state for requested type (donations or needs) and state for each type of asset fetched from the backend, be it needs or donations. Parsed out the requested asset on the BE and display those assets on the FE.

## Test Plan

repro steps:

1. Use Postman to make a post request to http://localhost:3000/api/assets. The BE requires the asset have a title, description, type, condition, and quantity.
image
2. From home/localhost:3000 search for assets by title. 
3. Assets will appear in the "Test" section at the bottom without an image.

## Resources


client/src/features/home/NeedsAndOffers.tsx
```diff
@@ -65,9 +65,17 @@ function NeedsAndOffers(props: Props): JSX.Element {
         {assets.map((asset) => (
           <NavLink to={`/asset/${asset.id}`} key={asset.id} className={classes.card}>
             <Card variant="outlined">
-              <img src={asset.imgUrls[0]} className={classes.cardImg} alt={asset.title} />
+              <img
+                src={
+                  asset.imgUrls
```
jd2rogers2, line 70: right idea, but i think it's probably better to do this a little higher up
`{asset.imgUrls.length && (<img ...`
then we don't need the fallback img you hardcoded here

client/src/features/action/assets/Assets.tsx
```diff
@@ -78,8 +78,13 @@ function Assets(): JSX.Element {
     });
   };
 
+  const [donations, setDonations] = React.useState([]);
```
jd2rogers2, line 81: i'm hoping this page will have a better design in the future to show just donations or just requests but for now let's implement getting both
so i'd say let's have 2 states, 2 fetches and each fetch will use a url like `/api/assets?type=request&search=${searchText}`
then on BE you can parse those query params via `@Query()` similar to `@Get()` in the transactions controller and select just a subset of the assets with each fetch

jd2rogers2 (commented):
on the right path here!
left a couple of tips but otherwise looks really good
keep it up!

client/src/features/action/assets/Assets.tsx
bidiaz101: Your tip to look at the `@Query()` in the transaction controller helped a lot, I think. I made another commit where I did my best to parse that out. Let me know what you think.

client/src/features/home/NeedsAndOffers.tsx
bidiaz101: When you say to do this higher up, you mean in the parent component? I'm not sure what your code snippet with `asset.imgUrls.length` is hinting at.

jd2rogers2 (commented):
so typically you'd move this out of draft state once it's beyond WIP/concept. so now that you're ready to receive feedback i'd say we're at that point
also, with that, please fill out the PR summary/description section with repro steps so that i can follow along and test on my side
(should include steps for creating an asset in the DB and then hitting FE path that displays that asset in the page)

bidiaz101:
> so typically you'd move this out of draft state once it's beyond WIP/concept. so now that you're ready to receive feedback i'd say we're at that point also, with that, please fill out the PR summary/description section with repro steps so that i can follow along and test on my side (should include steps for creating an asset in the DB and then hitting FE path that displays that asset in the page)

server/src/assets/assets.controller.ts
```diff
@@ -34,9 +36,12 @@ export class AssetsController {
...
 
   @Get()
-  findAll(): Promise<Asset[]> {
-    return this.assetsService.findAll();
+  get(@Query() getAssetsDto: GetAssetsDto): Promise<Asset[]> {
+    return this.assetsService.getAssets(getAssetsDto);
   }
+  // findAll(): Promise<Asset[]> {
+  //   return this.assetsService.findAll();
+  // }
```
jd2rogers2, line 44: remove

client/src/features/action/assets/Assets.tsx
```diff
@@ -78,9 +79,24 @@ function Assets(): JSX.Element {
...
+    'donation',
+  );
+  const [donations, setDonations] = React.useState<Asset[]>([]);
+  const [needs, setNeeds] = React.useState<Asset[]>([]);
+
   React.useEffect(() => {
     // fetch assets with querySearchText
-  }, [location]);
+    fetch(`http://localhost:3001/api/assets?type=${selectedAssetType}&title=${querySearchText}`)
+      .then((resp) => resp.json())
```
jd2rogers2, line 91: looks like filtering on `title` is blocking this request
why don't we update the url to `fetch('blah?type={}${querySearchText ? '&title=${querySearchText}' : ''`}'`

client/src/features/action/assets/Assets.tsx
```diff
@@ -133,6 +151,10 @@ function Assets(): JSX.Element {
           </Paper>
           <NeedsAndOffers headerText="Nonprofit Needs" assets={mockData} />
           <NeedsAndOffers headerText="Offers" assets={mockData} />
+          <NeedsAndOffers
+            headerText="Test"
+            assets={selectedAssetType === 'donation' ? donations : needs}
+          />
```
jd2rogers2, line 157: let's just remove the Offers section altogether and make yours have `headerText="Offers"`

client/src/features/home/NeedsAndOffers.tsx
```diff
@@ -65,9 +65,11 @@ function NeedsAndOffers(props: Props): JSX.Element {
         {assets.map((asset) => (
           <NavLink to={`/asset/${asset.id}`} key={asset.id} className={classes.card}>
             <Card variant="outlined">
-              <img src={asset.imgUrls[0]} className={classes.cardImg} alt={asset.title} />
+              {asset.imgUrls ? (
+                <img src={asset.imgUrls[0]} className={classes.cardImg} alt={asset.title} />
+              ) : null}
```
jd2rogers2, line 70: hmm, maybe i take it back....
go with the default image solution you had before

jd2rogers2 (approved):
couple small nits but then it's ready to 🚀🚀🚀🚀🚀🚀

server/src/assets/assets.controller.ts
bidiaz101: oops! Thanks! 
Do we still need the findAll() method in assets.service.ts if it's not used here? Is there any downside to leaving it there in case users do want to see all assets?

client/src/features/action/assets/Assets.tsx
bidiaz101: I made the change, but I not sure why. Is there somewhere on the backend I should look to understand what happening there?

server/src/assets/assets.controller.ts
jd2rogers2: we should remove any unused code. less future wasted work maintaining something that's not being used

client/src/features/action/assets/Assets.tsx
jd2rogers2: yep, everything after the `?` in the url is the query params. so `title` finds its way into `@Query()` and the `getAssetsDto` which is what filters the assets list
happy to discuss that further if needed

client/src/features/action/assets/Assets.tsx
```diff
@@ -131,8 +153,11 @@ function Assets(): JSX.Element {
               </Button>
             </NavLink>
           </Paper>
-          <NeedsAndOffers headerText="Nonprofit Needs" assets={mockData} />
-          <NeedsAndOffers headerText="Offers" assets={mockData} />
+          <NeedsAndOffers headerText="Nonprofit Needs" assets={mockData} />\
```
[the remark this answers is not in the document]
bidiaz101: Thanks! Removed the slash and extra findAll() method in assets.service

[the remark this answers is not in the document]
bidiaz101: Yes please! I'm assuming github's instructions won't work since the main branch is protected?
image

Issue #118, opened 2021-12-27
interrupt page refresh or page exit if form is not blank


jd2rogers2:
a user fills out a form, then tries to leave without submitting
we want to interrupt this and check that they are doing so intentionally

i'm not sure how to do this TBH so research is required here

paaaarv:
i'd like to pick this ticket up! just a thought though -- could a function that maps through state to see if anything doesn't equal " " before leaving? and then can pop up an alert?

jd2rogers2:
> i'd like to pick this ticket up! just a thought though -- could a function that maps through state to see if anything doesn't equal " " before leaving work? and then can pop up an alert? 

Go for it! Yep, pretty much! 
Idk if map will work but I think your idea is going in the right direction. And just to be clear, not a native js alert but a material ui/home cooked modal

And a little more direction here: 
- Just get it working in one of our forms to start and then in a follow up PR we can reuse in the other places
- not sure exactly how it'll work but there will essentially be some sort of listener/callback, and in that call back you should use a function that takes in the form state (object with any keys so that it can be reused for other forms) to check the form as you described

PR #149, opened 2022-03-09
Implemented custom Alert Dialog on route change

ObieMunoz:
I created a new shared component and utility, AlertDialog and DetectFormData, which can be integrated into pages with formData to interrupt the user from exiting if a form has been partially filled out.

This is currently implemented in NeedFormGoods and can be implemented in other forms app-wide.

Resolves issue #118 

Strongly referenced Yadhunandan's guide here https://medium.com/nerd-for-tech/custom-react-router-prompt-d325538b4d2b


client/src/assets/sharedComponents/DetectFormData.tsx
```diff
@@ -0,0 +1,10 @@
+function DetectFormData(formData: object) {
+  for (const key in formData) {
+    if ((formData as any)[key] !== '') {
+      return true;
+    }
+  }
+  return false;
+}
```
jd2rogers2, line 8: would something a little simpler and more readable work
like:
`Object.values(formData).filter(val => Boolean(val)).length > 0`
haha better readability is debatable now that i've written it out
i'll still leave it here and leave it up to you

client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -55,8 +57,13 @@ const initialFormData: ShareANeedData = {
 
 function NeedForm() {
   const [formData, setFormData] = React.useState<ShareANeedData>(initialFormData);
+  const [showPrompt, setShowPrompt] = React.useState(false);
   const history = useHistory();
 
+  React.useEffect(() => {
+    setShowPrompt(() => DetectFormData(formData));
+  }, [showPrompt, formData]);
```
jd2rogers2, line 65: don't think we want `showPrompt` in the dependencies here
that could lead to infinite loop

client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -55,8 +57,13 @@ const initialFormData: ShareANeedData = {
 
 function NeedForm() {
   const [formData, setFormData] = React.useState<ShareANeedData>(initialFormData);
+  const [showPrompt, setShowPrompt] = React.useState(false);
```
jd2rogers2, line 60: let's type the state `React.useState<boolean>(false);`

jd2rogers2: this state doesn't define whether we're showing the prompt so i'd say naming should be something more like `isFormStarted` or `isFormInProgress`

client/src/assets/sharedComponents/AlertDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const [showPrompt, setShowPrompt] = React.useState(false);
+  const [selectedRoute, setSelectedRoute] = React.useState('');
+
+  useEffect(() => {
+    if (when) {
+      history.block((prompt) => {
+        setSelectedRoute(prompt.pathname);
+        setShowPrompt(true);
+        return 'true';
```
jd2rogers2, line 23: is this meant to be stringified? what's this return value doing?
remember to post any articles, documentation you used in your PR description plz

jd2rogers2 (commented):
looking pretty good!
couple small notes, but should be quick approve on next round

client/src/assets/sharedComponents/AlertDialog.tsx
ObieMunoz: It's intended, history.block will throw an error here if it's a boolean.

client/src/assets/sharedComponents/DetectFormData.tsx
ObieMunoz: yeah it's going to look weird both ways

client/src/features/action/needs/NeedFormGoods.tsx
ObieMunoz: updated

client/src/features/action/needs/NeedFormGoods.tsx
ObieMunoz: updated to isFormInProgress

client/src/assets/sharedComponents/AlertDialog.tsx
ObieMunoz: updated description

client/src/features/action/needs/NeedFormGoods.tsx
```diff
@@ -55,8 +57,13 @@ const initialFormData: ShareANeedData = {
 
 function NeedForm() {
   const [formData, setFormData] = React.useState<ShareANeedData>(initialFormData);
+  const [formInProgress, setFormInProgress] = React.useState<boolean>(false);
   const history = useHistory();
 
+  React.useEffect(() => {
+    setFormInProgress(() => DetectFormData(formData));
```
jd2rogers2, line 64: any specific reason for the arrow function?
or would `setFormInProgress(DetectFormData(formData));` work?

jd2rogers2 (approved):
did you play around with https://stackoverflow.com/questions/821011/prevent-a-webpage-from-navigating-away-using-javascript at all?
not saying we want to do this (design/product question) but just curious if it works

client/src/features/action/needs/NeedFormGoods.tsx
ObieMunoz: I don't ever set state in that manner, it can lead to unintended results in certain scenarios.

PR #150, opened 2022-03-16
Fix login, update node, fix husky commands

jd2rogers2:
## Dev Summary

bug with login is fixed
bug was caused by excluding password property at the entity level, but that is needed in the check password auth flow
solution was to remove entity exclusion code and delete password at earliest point in all endpoints

also:
updates node version because i was getting an error when committing 🤷  we should update anyways
updates husky commands so we don't increment dependencies so often. causes bloat in PRs and occasional local env issues

## Test Plan

repro steps:

1. `nvm use` then start local process
2. localhost:3000/login
3. login with existing user
expect success

Screen Shot 2022-03-19 at 11 46 48 AM

## Resources

- NA


server/src/auth/auth.service.ts
```diff
@@ -10,35 +10,26 @@ export class AuthService {
...
-          status: HttpStatus.UNAUTHORIZED,
-          error: 'Invalid password',
-        },
-        HttpStatus.UNAUTHORIZED,
-      );
-    }
   }
 
-  async createJwt(user: User) {
-    delete user.password;
```
jd2rogers2, line 41: i checked, pw already excluded

server/src/auth/strategies/login.strategy.ts
```diff
@@ -13,9 +13,6 @@ export class LoginStrategy extends PassportStrategy(Strategy) {
 
   async validate(email: string, password: string): Promise<any> {
     const user = await this.authService.validateUser(email, password);
-    if (!user) {
-      throw new UnauthorizedException();
-    }
```
jd2rogers2, line 18: error already thrown in auth service

server/src/users/entities/user.entity.ts
```diff
@@ -19,7 +19,7 @@ export class User {
   @Column({ type: 'text', unique: true })
   email: string;
 
-  @Column('text', { select: false })
```
jd2rogers2, line 22: adding this was what broke things
pw was being excluded in the user service select/get from db
therefore we couldn't check if pw was correct in auth service

server/src/users/users.service.ts
```diff
@@ -26,34 +29,44 @@ export class UsersService {
...
   }
 
-  //TODO: Assess if there is a better way than making two requests.
+  // TODO: Assess if there is a better way than making two requests.
   async update(id: number, updateUserDto: UpdateUserDto) {
     await this.usersRepository.update(id, updateUserDto);
-    return this.usersRepository.findOne(id);
+    const user = await this.usersRepository.findOne(id);
+    delete user.password;
+    return user;
```
jd2rogers2, line 66: now deleting pw at earliest point in all cases
(aka in here except for 1 case where we need to check pw in auth service. in that case we delete right after pw check)

.nvmrc
```diff
@@ -1 +1 @@
-14.16.0
\ No newline at end of file
+14.18.3
```
esteban-gs, line 1: Should we use `16.14.2`? it's the LTS. If we do, it should probably be on another PR.

esteban-gs:
Looks good! Can you try to remove the `skip` part and see if you can get it to pass?

https://github.com/Nonprofit-Exchange-Hub/web-app/blob/fc08972728af5d7d00cbe2253605abfd39587989/server/test/users/users.controller.e2e-spec.ts#L113-L123

.nvmrc
jd2rogers2: good call, put in a ticket for it
https://github.com/Nonprofit-Exchange-Hub/web-app/issues/151
still going to bump up in this PR if that's ok

jd2rogers2:
> Looks good! Can you try to remove the `skip` part and see if you can get it to pass?
Done!! good call
ready for another review round

server/test/users/users.controller.e2e-spec.ts
```diff
@@ -110,7 +110,7 @@ describe('UsersController', () => {
     });
 
     // not yet implemented, so skipping for now
```
esteban-gs, line 112: remove comment

esteban-gs:
Looks good to me. Just remove the comment and merge

esteban-gs (approved):

PR #153, opened 2022-03-28
get users transactions

jd2rogers2:
## Dev Summary

we want to see a user's transactions on the messages inbox page
- create a fetch to get these
- create endpoint that uses user cookie to get transactions
- needed to include a user's organization in user service get user

also:
- added is loading state to user context provider and consumed it in inbox page
  - (so we don't hit the redirect to home when no user logic below the loading return)
- added more properties to FE types
- add `@UseGuards(CookieAuthGuard)` to transactions endpoints so only logged in users can see them
  - (will need to later add checks for authorization (user is involved with transaction, otherwise shouldn't be able to see it))
- simplified user<>org relationship. deleted userOrg table (which we were using to house the 'role' property) in lieu of a permissions table (not impl here, will be in future PR)
- lots of cleanup
  - renamed `last_name` to `lastName`
  - fixed 'justifyContent' lint

## Test Plan

repro steps:

1. confirm that a user can see message threads (transactions) about assets they posted
1a. create a user
```
curl -X POST localhost:3001/api/users \
-H "Content-Type: application/json" \
-d '{ "firstName": "test", "lastName": "1", "password": "test", "email": "test1@test.com" }'
```
1b. create an asset that above user posted
```
curl -X POST localhost:3001/api/assets \
-H "Content-Type: application/json" \
-d '{ "title": "thing 1", "description": "its a thing", "type": "donation", "quantity": "1",
"poster": { "id": "1" } }'
```
1c. create an org
```
curl -X POST localhost:3001/api/organizations \
-H "Content-Type: application/json" -d '{
  "name": "name",
  "description": "description",
  "website": "website.com",
  "address": "address",
  "phone": "phone",
  "city": "seattle",
  "state": "wa",
  "ein": "1",
  "tax_exempt_id": "1"
}'
```
1d. need to comment out `UseGuards` line above `@Post()` in transactions controller
1e. create the transaction
```
curl -X POST localhost:3001/api/transactions \
-H "Content-Type: application/json" \
-d '{ "poster": { "id": "1" }, "recipient": { "id": "1"}, "asset": { "id": "1"} }'
```
1f. localhost:3000 > log in to that user > go to localhost:3000/inbox > see transaction
see network tab with `/transactions/current-user` request that responds with a transactions array
see transaction title displayed in message threads list (left sidebar)
Screen Shot 2022-04-12 at 9 39 17 AM

2. user can see their org's transactions
2a. create user with org
```
curl -X POST localhost:3001/api/users \
-H "Content-Type: application/json" \
-d '{ "firstName": "test", "lastName": "3", "password": "test", "email": "test3@test.com", "organizations": [{ "id": "1" }] }'
```
2b. localhost:3000 > log in to that user > go to localhost:3000/inbox > see transaction

3. confirm other users cannot see others' transactions
3a. create user
```
curl -X POST localhost:3001/api/users \
-H "Content-Type: application/json" \
-d '{ "firstName": "test", "lastName": "2", "password": "test", "email": "test2@test.com" }'
```
3b. localhost:3000 > log in to that user > go to localhost:3000/inbox > see 0 transactions

## Resources

- https://rubular.com/r/BDDaaHGBSoHo5O


client/src/providers/UserProvider.tsx
```diff
@@ -56,5 +61,9 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
     setUserTimeout(null, true, true);
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
-  return <UserContext.Provider value={[user, setUserTimeout]}>{children}</UserContext.Provider>;
+  return (
+    <UserContext.Provider value={[user, setUserTimeout, isLoading]}>
+      {children}
+    </UserContext.Provider>
+  );
```
jd2rogers2, line 68: added loading state so that we can handle with `if isLoading return null` in our components and therefore always have `user` var in renders and not have to deal with conditionals

server/src/assets/entities/asset.entity.ts
```diff
@@ -1,8 +1,9 @@
 import { Entity, Column, PrimaryGeneratedColumn, ManyToOne, JoinColumn, OneToMany } from 'typeorm';
-import { User } from '../../users/entities/user.entity';
 
-import { AssetType, Condition } from '../constants';
+import { User } from '../../users/entities/user.entity';
```
esteban-gs, line 3: Nitpick:
Optional, remove unused import

server/src/organizations/entities/organization.entity.ts
```diff
@@ -40,7 +41,10 @@ export class Organization {
 
   @OneToMany(
     () => Transaction,
-    (transaction) => transaction.donater_organization || transaction.recipient,
+    (transaction) => transaction.asset.posterOrganization || transaction.recipient,
```
esteban-gs, line 44: So, is this (`transaction.asset.posterOrganization || transaction.recipient`) saying that it could have a relationship to either the asset or the transactions tables at anytime?

server/src/transactions/transactions.controller.ts
```diff
@@ -1,40 +1,72 @@
...
 
+  @UseGuards(CookieAuthGuard)
   @Get()
-  get(@Query() getTransactionsDto: GetTransactionsDto): Promise<Transaction[]> {
+  async get(@Query() getTransactionsDto: GetTransactionsDto): Promise<Transaction[]> {
     return this.transactionsService.getTransactions(getTransactionsDto);
   }
 
+  @UseGuards(CookieAuthGuard)
+  @Get('/currentUser')
```
esteban-gs, line 45: Optional:
- web routes don't differentiate between `currentuser` or `currentUser`, if you want, change it to: `current-user`.

server/src/transactions/transactions.service.ts
```diff
@@ -30,6 +32,16 @@ export class TransactionsService {
     return found;
   }
 
+  // TODO make omit user its own type
+  async getCurrentUsersTransactions(user: Omit<User, 'password'>): Promise<Transaction[]> {
+    const found = await this.transactionsRepository.find({
+      where: [{ recipient: user.organizations }, { asset: user.assets }],
```
esteban-gs, line 38: I think you need to first get the assets belonging to this user, then use the asset ID to find its transactions

esteban-gs (changes requested):
Awesome! I was able to reproduce your feature fairly easily. 

Except for a couple of things to take a closer look at it should be ready.

server/src/transactions/transactions.controller.ts
jd2rogers2: actually going to opt to just use the existing `Get('/')` endpoint and pass user_id in as part of the query params

client/src/features/action/assets/Asset.tsx
```diff
@@ -154,14 +154,14 @@ function Asset(): JSX.Element {
     });
 
   const aboutInfo = () => {
-    if (asset.organization !== null) {
```
jd2rogers2, line 157: asset shouldn't have an organization
just put whatever in here since this will all probs get re-written in FE re-write initiative

client/src/features/users/Inbox/Inbox.tsx
```diff
@@ -153,21 +126,26 @@ function MessageInboxView(): JSX.Element {
     }
   }, [selectedTransaction]);
 
+  if (isLoading) {
+    return null;
+  }
+
```
jd2rogers2, line 132: needed this here to prevent us from hitting redirect below

server/src/organizations/entities/organization.entity.ts
jd2rogers2: > So, is this (`transaction.asset.posterOrganization || transaction.recipient`) saying that it could have a relationship to either the asset or the transactions tables at anytime?
it's saying that an org can have transactions via an org's employees posting or via claiming an asset. so i think that's a yes to your question?

> This, I believe creates a relationship from `Org` to `Transactions` via transactions.recepientId
correct, i think that's what we want

server/src/transactions/transactions.controller.ts
jd2rogers2: nvm, went with current-user

server/src/transactions/transactions.service.ts
jd2rogers2: reworked this a bit, lmk what you think. tested and confirmed that user not included in transaction cannot see that transaction (and included in test plan)

server/src/transactions/entities/transaction.entity.ts
```diff
@@ -44,7 +36,16 @@ export class Transaction {
   @JoinColumn()
   recipient: Organization;
 
-  @OneToMany(() => Message, (message) => message.transaction)
+  @ManyToOne(() => User, (poster) => poster.transactions)
   @JoinColumn()
+  poster: User;
```
jd2rogers2, line 41: TODO rename to something like 'donater'

server/src/organizations/entities/organization.entity.ts
```diff
@@ -32,15 +33,15 @@ export class Organization {
...
-
-  @OneToMany(() => UserOrganization, (user_org) => user_org.organization)
-  users: UserOrganization[];
+  @ManyToMany(() => Organization, (org) => org.users)
+  users: User[];
 
   @OneToMany(
     () => Transaction,
-    (transaction) => transaction.donater_organization || transaction.recipient,
+    (transaction) => transaction.asset.posterOrganization || transaction.recipient,
```
jd2rogers2, line 41: TODO test that this relationship is working

server/src/transactions/transactions.service.ts
```diff
@@ -30,6 +32,22 @@ export class TransactionsService {
     return found;
   }
 
+  async getCurrentUsersTransactions(user: Omit<User, 'password'>): Promise<Transaction[]> {
+    if (!user.assets?.length && !user.organizations?.length) {
+      return [];
+    }
+
+    const orgFilters = user.organizations.map((org) => ({ recipient: { id: org.id } }));
+    const assetFilters = user.assets.map((asset) => ({ asset: { id: asset.id } }));
```
jd2rogers2, line 41: TODO does this get the transactions that the user's org has donated??

PR #156, opened 2022-04-06
Integrate create org

esteban-gs:
## Dev Summary
- Created types for form data
- Added `React Hook Form` and `Yup` for form logic and form validation to `client/package*`
- Added `react-query` and `axios` to deal with multiple interdependent api requests and errors
- Removed manual validation methods and replaced with `Yup`
- Expanded `TextField` wrapper component props
- Added logic for error messages

## Test Plan
repro steps:
Enter info for non-existing orgs or users
1. go to http://localhost:3000/signup-nonprofit
2. Fill ein in  step 1
3. Fill out org info in  step 2
4. Fill out user info in step 3
5. click submit
6. expect to see a success notification
7. Click submit again and expect to see a notification that this user already exists

#### Expect
[See my video demo](https://www.loom.com/share/a96766ed4f5f407e96f65f834c170686)

## Resources

- [simple validation](https://thewebdev.info/2022/01/06/how-to-invalidate-a-textfield-in-react-material-ui/)
- [formik/gulp ways](https://formik.org/docs/examples/with-material-ui)
- [Formik docs](https://formik.org/docs/overview)
- [Yup Docs](https://www.npmjs.com/package/yup)


client/src/types/index.ts
```diff
@@ -10,6 +11,59 @@ export type Asset = {
...
+  email: string;
+};
+
+export type UserEntityCreateObj = {
+  firstName: string;
+  last_name: string;
+  email: string;
+  password: string;
+};
+
```
jd2rogers2, line 66: all this redundancy is not good. let's have:
```
type BaseUser = {
  firstName: string;
  last_name: string;
  email: string;
};

type User = BaseUser & { id: string };

type NewUser = BaseUser & { password: string };
```
or something along these lines

server/src/user-org/user-org.service.ts
```diff
@@ -15,6 +15,7 @@ export class UserOrganizationsService {
   ) {}
 
   async create(createUserOrganizationDto: CreateUserOrganizationDto): Promise<UserOrganization> {
+    console.log(createUserOrganizationDto, 'dot from userorg');
```
jd2rogers2, line 18: remove log unless you are intentionally keeping it in

server/src/organizations/organizations.controller.ts
```diff
@@ -22,6 +22,7 @@ export class OrganizationsController {
 
   @Post()
   async create(@Body() createOrganizationDto: CreateOrganizationDto): Promise<Organization> {
+    console.log(createOrganizationDto, 'ORG');
```
jd2rogers2, line 25: remove log

client/src/types/index.ts
```diff
@@ -10,6 +11,59 @@ export type Asset = {
...
+  state: string;
+  ein: string;
+  nonprofit_classification: string;
+};
+
+export type UserOrg = {
+  id?: number;
+  approvalStatus: ApprovalStatus;
+  role: Role;
+};
```
jd2rogers2, line 32: 🙃 i'm removing userOrg type in my PR
lmk if this effects your work here. if so we can look at conflicts together and see what best route is

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+    AxiosResponse<any, any>,
+    unknown,
+    { ein: string; name: string },
+    string[]
+  >({
+    enabled: triggerEinSearch,
+    queryKey: ['orgValidateEinQuery', org.ein],
+    queryFn: ({ queryKey }) => {
+      const [_key, ein] = queryKey;
+      console.log(_key); // need to use _key somehow
```
jd2rogers2, line 77: i think you can destructure like `[, ein] =`

jd2rogers2 (commented):
partial review here because i'm getting interrupted
but a couple notes for you to go off of for now at least

client/src/types/index.ts
esteban-gs: Thanks! that is a great point. I moved one to the one place it is used and made it local. I refactored to other 2 to avoid repetition.

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
esteban-gs: Nice. It worked

server/src/user-org/user-org.service.ts
esteban-gs: I removed this

server/src/organizations/organizations.controller.ts
esteban-gs: I removed this

client/src/features/users/Auth/SignupNonProfit.tsx
```diff
@@ -43,84 +32,87 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   last_name: '',
-  role_or_title: '',
   email: '',
-  password: '',
-  accept_terms: false,
+};
+
+const defaultUserOrg: UserOrg = {
+  approvalStatus: ApprovalStatus.pending,
+  role: Role.owner,
```
jd2rogers2, line 56: i think if someone is signing up a new nonprofit we should just make them the org owner right off the bat. also their approvalStatus is kind of moot at that point
so i think we can remove this state and set these things by hardcoding o default on BE

client/src/features/users/Auth/SignupNonProfit.tsx
```diff
@@ -43,84 +32,87 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-  const handleSelectChange = (evt: React.ChangeEvent<HTMLSelectElement>) => {
-    const { name, value } = evt.target;
-
-    setFormData((fData) => {
-      return {
-        ...fData,
-        [name]: value,
-      };
-    });
+  const onChildSetParentUserOrg = (userOrg: UserOrg): void => {
```
jd2rogers2, line 79: hmm i'm not understanding the "parent" and "child" terms here?

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+          {orgCreateMutation.isError && (
+            <SimpleSnackbar message={`An error occurred ${orgCreateMutation.error}`} />
+          )}
+
+          {orgCreateMutation.isSuccess && <SimpleSnackbar message={`Organization created`} />}
+        </>
+      )}
+      <Formik initialValues={org} validationSchema={SignupSchema} onSubmit={() => {}}>
+        {({ handleChange, values, touched, errors, setFieldTouched, setFieldValue, isValid }) => (
+          <form>
```
jd2rogers2, line 113: my thinking is that we shouldn't create an org and then create a user separately
this current pattern allows for just the org (with no user) to be created
so i'd think these 2 pages/steps should be wrapped by 1 `<form>` element that sends 2 simultaneous requests or hits 1 endpoint with all the data and that endpoint handles the 2 db writes
thoughts?

client/src/types/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
jd2rogers2, line 1: i think there's a version that it will disable the rule just for 1 line rather than the entire file
i prefer this so that the code is closer by and more easily associated to the unused var
that way we don't delete/use the var and forget to remove this

client/src/types/index.ts
jd2rogers2: on second thought, let's try to get yours landed and then i can do mine after
but still with this entity, it should be created under the hood i.e. FE doesn't know anything about it
pretty sure this will be possible if we do the 1 form pattern i mentioned

jd2rogers2 (commented):
loving the new libraries (still researching them tho)
and looks pretty good overall for the most part
1 preference on the single form idea
otherwise only small comments here

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+const classifications = [
+  { value: 'charitable', text: 'Charitable Organization' },
+  { value: 'religious', text: 'Religious Organization' },
+  { value: 'private', text: 'Private Foundation' },
+  { value: 'political', text: 'Political Organizations' },
+  { value: 'other', text: 'Other' },
+];
+
+const SignupSchema = Yup.object().shape({
```
jd2rogers2, line 23: let's rename so these schemas don't share a name

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+    .matches(/^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]\d{3}[\s.-]\d{4}$/, 'EIN must match: 99-9999999')
+    .required('Required'),
+  nonprofit_classification: Yup.string().min(2).required('Required'),
+});
+
+interface Props {
+  triggerNextStep: (step: number) => void;
+  setParentOrg: (createdOrg: Organization) => void;
+  parentOrg: Organization;
+  classes: ClassNameMap<any>;
```
jd2rogers2, line 44: hmm, not liking the `any` but yeah looking at the source code and it's not pretty
so this might have to do
can we leave a comment saying this is essentially what the source code has this typed as

jd2rogers2 (commented):
finished researching the new pkgs and 👍 
couple more nits

client/src/features/users/Auth/SignupNonProfit.tsx
esteban-gs: Parent <-> child component communication

esteban-gs:
TODO:
- validate org in db with EIN check
- Use one form instead of two

client/src/features/users/Auth/SignupNonProfit.tsx
jd2rogers2: ok i think with the new pattern there will only be 1 state so we shouldn't really need the parent/child labeling?

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+  state: Yup.string().required('Required').min(2).max(2),
+  ein: Yup.string()
+    .matches(/^[0-9]\d?-\d{7}$/, 'EIN must match: 99-9999999')
+    .required('Required'),
+  name: Yup.string(),
+  description: Yup.string().required('Required'),
+  website: Yup.string().required('Required'),
+  address: Yup.string().required('Required'),
+  phone: Yup.string()
+    .matches(/^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]\d{3}[\s.-]\d{4}$/, 'EIN must match: 99-9999999')
```
jd2rogers2, line 41: fix error message

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+    onError: (res: any) => {
+      setOrg({ ...org, name: '' });
+    },
+    retry: 0,
+  });
+
+  const handleNext = (values: Organization) => {
+    if (org.id === undefined) {
+      setOrg({ ...org });
+      orgCreateMutation.mutate({ ...values, name: org.name });
```
jd2rogers2, line 100: seems to still be making request to create a form on `next` click from step 1

Screen Shot 2022-06-17 at 21 49 59

jd2rogers2: i think `handleNext` should only do `triggerNextStep(1);` right?
well, and validate that all inputs are filled out

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+const CreateOrgForm = ({
+  setParentOrg = () => {},
+  parentOrg,
+  triggerNextStep = () => {},
+  classes,
+}: Props) => {
+  const [org, setOrg] = React.useState<Organization>(parentOrg);
+  const [triggerEinSearch, setTriggerEinSearch] = React.useState<boolean>(false);
+
+  const orgCreateMutation = useMutation<AxiosResponse<any, any>, Error, Organization, Error>(
```
jd2rogers2, line 61: this mutation should be moved to parent and be called at same time (or right after) user create request

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+  parentOrg: Organization;
+  classes: ClassNameMap<'button' | 'header' | 'arrow' | 'sideImg' | 'signUpContainer'>;
+}
+const CreateOrgForm = ({
+  setParentOrg = () => {},
+  parentOrg,
+  triggerNextStep = () => {},
+  classes,
+}: Props) => {
+  const [org, setOrg] = React.useState<Organization>(parentOrg);
```
jd2rogers2, line 58: duplicate state with `parentOrg` we should be able to have just `parentOrg`

jd2rogers2 (commented):
saw some changes were made but maybe this isn't ready??
left some comments anyways

client/src/features/users/Auth/Steps/CreateOrgForm.tsx
esteban-gs: I removed the need to pass it to a child component, so it is no longer necessary

esteban-gs:
@jd2rogers2 , I think I have addressed all your review items. I also added a test for the org create logic. Want  to take another look?

client/src/types/index.ts
esteban-gs: Thanks, I moved everything into 1 form

jd2rogers2 (approved):
it works!

some small changes needed though
- [ ] in line comments
- [ ] some server side start errors. i.e.
```
[7:49:32 AM] Starting compilation in watch mode...

src/database/seeding/seed-data.ts:100:5 - error TS2741: Property 'nonprofit_classification' is missing in type '{ name: string; doing_business_as: string; description: string; website: string; address: string; phone: string; city: string; state: string; ein: string; }' but required in type 'CreateOrganizationDto'.

100     {
        ~
101       name: 'Childrens Home Society of Washington',
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
109       ein: '12-3456788',
    ~~~~~~~~~~~~~~~~~~~~~~~~
110     },
    ~~~~~
```

non blocking (so maybe follow up PR(s)):
- you mentioned maybe a better context provider pattern you wanted to try with formik
- separate display components for each tab and then pass formik and states as props
- you also made a good call of a possible `SignUpDataWrapper` parent component that houses all the state

server/src/organizations/organizations.service.ts
```diff
@@ -46,6 +71,10 @@ export class OrganizationsService {
     return this.organizationsRepository.findOne(id);
   }
 
+  countByNameOrEin(name: string, ein: string): Promise<number> {
+    return this.organizationsRepository.count({ name, ein });
+  }
```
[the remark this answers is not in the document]
esteban-gs: > i'm not seeing documentation for `.count`. are you sure this is an `or` situation and it's not counting orgs with matching name `and` ein?

doc for count is [HERE](https://typeorm.io/repository-api)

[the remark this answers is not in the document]
esteban-gs: Changed it to `OR` based on [this](https://stackoverflow.com/a/62889991/13862254)

esteban-gs:
@jd2rogers2 , I switched from Formik to react-hook-forms. It's a lot easier to use. I also cleaned up some of the extraneous things to separate modules.

PR #157, opened 2022-04-24
added location to assets BE & FE

bidiaz101:
## Dev Summary

[Mocks](link)

Added two columns to asset entity: location and date posted. Location is null until Design decides how they'd like that to be posted (via user input or location feature). Date posted is formatted client-side with new Date().toLocaleDateString()

## Test Plan

repro steps:

1. Use Postman to post an asset to backend/api/assets
2. Format raw JSON as I did below. I found out poster object needs a minimum of an id to be valid. JSON on the bottom is the response with the date posted added by the backend.
image
4. Navigate to client homepage (localhost:3000) 
5. Search for assets by title
6. Assets will appear with the date they were posted
expected: 

image

## Resources

### DIFF
--- a/client/src/features/action/assets/Asset.tsx
+++ b/client/src/features/action/assets/Asset.tsx
@@ -137,21 +137,22 @@ function Asset(): JSX.Element {
     return <>asset not found</>;
   }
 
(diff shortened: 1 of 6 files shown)


jd2rogers2 (approved):
nice!!

PR #158, opened 2022-04-27
FE: Password reset flow

bidiaz101:
## Dev Summary

Created new component for password reset. In progress.

## Test Plan

repro steps:

1. localhost:3000/set-new-password
2. See new component

## Resources

### DIFF
--- a/client/src/features/users/Auth/SetNewPassword.tsx
+++ b/client/src/features/users/Auth/SetNewPassword.tsx
@@ -1,0 +1,11 @@
+import * as React from 'react';
+
+function SetNewPassword() {
+  return (
+    <div>
+      <h1>SetNewPassword</h1>
+    </div>
+  );
+}
+
+export default SetNewPassword;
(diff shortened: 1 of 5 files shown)


client/src/features/users/Auth/SetNewPassword.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const handleChange = (evt: React.ChangeEvent<HTMLInputElement>): void => {
+    setPassword(evt.target.value);
+  };
+
+  const [user] = React.useContext(UserContext);
+
+  const handleSubmit = async (evt: React.FormEvent): Promise<void> => {
+    evt.preventDefault();
+    // ternary in url temporary. User should be logged in(?) id should be available once BE is built
+    fetch(`http://localhost:3001/api/users/${user ? user.id : null}`, {
```
esteban-gs, line 54: Add the property to include the credentials with the request. See this as a reference: 
https://github.com/Nonprofit-Exchange-Hub/web-app/blob/b7ba0f60b77a889adbbf21755ea5fcd59b1acb0e/client/src/providers/UserProvider.tsx#L22-L27

esteban-gs (changes requested):
Very good! See the comment about including credentials.

client/src/features/users/Auth/SetNewPassword.tsx
bidiaz101: Thanks! Made the change. I will definitely be looking into sending a fetch with credentials and how that works.

esteban-gs (approved):

PR #192, opened 2022-06-21
remove use of mockData (assets) and replace with BE data, other small fixes

jd2rogers2:
## Dev Summary

frontend has been using hardcoded data. this uses data from the db and adds seed data

## Test Plan

repro steps:

see BE data
1. from server, before starting, seed new assets `npm run seed:drop`
2. start servers
3. localhost:3000
4. open dev tools network tab
see that we made 2 calls to /assets and that data is what we're seeing as asset cards
5. click on one of the asset cards
6. see asset show page that matches card

see FE routing is fixed
1. localhost:3000
2. click `make a post` button
3. click `goods`

## Resources

- [orm find function](https://typeorm.io/find-options)


server/src/assets/dto/get-asset.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { PartialType } from '@nestjs/mapped-types';
```
jd2rogers2, line 1: file renamed from `server/src/assets/dto/get-asset-filter.dto.ts`

server/src/assets/dto/get-asset.dto.ts
```diff
@@ -0,0 +1,12 @@
...
+import { IsOptional } from 'class-validator';
+
+import { Asset } from '../entities/asset.entity';
+
+export class GetAssetsDto extends PartialType(Asset) {
+  @IsOptional()
```
esteban-gs, line 7: Is the intention of extending from the entity that you can use the entity properties to filter?

jd2rogers2: huh...
yes, i believe so. but also, i was just following the pattern we've been doing. i don't remember why we started doing this, and now that i look, they aren't doing that in the nest docs
so i think creating less coupling to the entity would be better. will remove extensions

README.md
```diff
@@ -18,7 +18,9 @@
         * Confirm installation of Docker Compose: `$ docker-compose --version`
 2. Download or [clone](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository-from-github/cloning-a-repository) repository.
 3. Run `nvm use` to ensure you are using the proper node version
-4. Install required dependencies by running `npm install` from within `/server` directory and from the `/client` directory.
+4. Install required dependencies:
+  -  run `npm install` from within `/server` directory and from the `/client` directory.
+  - run `npm run prepare` from within `/server` directory and from the `/client` directory.
```
esteban-gs, line 23: Since you are already here, can you change it to `npm ci` instead of `npm install`

client/src/features/home/Home.tsx
```diff
@@ -161,6 +163,25 @@ function Home(): JSX.Element {
   const classes = useStyles();
   const [selectedSearchCategory, setSelectedSearchCategory] = React.useState<string>('');
   const [searchText, setSearchText] = React.useState<string>('');
+  const [donations, setDonations] = React.useState<Asset[]>([]);
```
esteban-gs, line 166: Recommended:
Can you use the javascript URL api to generate the url?
https://www.valentinog.com/blog/url/

esteban-gs (approved):
Nice! some small comments with recommended changes but not required

PR #193, opened 2022-06-24
lerna ize

jd2rogers2:
## Dev Summary
[lerna](https://lerna.js.org/docs/introduction) is a tool for managing monorepos
here, i have added lerna which will simplify our scripting and move shared node_modules into a root node_modules folder, making our package lighter

don't be intimidated by "files changed" count, it's mostly just moving files from /server to /packages/server etc.

## Test Plan

repro steps:

1. from root run `npm run bootstrap`
expect node_modules to be created in each package and in root
2. run `npm run prepare`
idk if this is actually needed but we have new husky commands and a new file structure so probs best to help husky configure and get it's bearings
3. run `npm run checks`
expect all tests to run and pass
4. run `npm run start`
expect both servers to start in same terminal
expect localhost:3000 to work in entirety

## Resources

- adding lerna https://www.youtube.com/watch?v=j0FiMekdeOs
- https://medium.com/shopback-tech-blog/5-tips-about-lerna-4186840093f2


.husky/post-checkout
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 FILE="$(dirname "$0")/_/husky.sh"
 if [ ! -f "$FILE" ]; then
-  cd server && npm ci && npm run prepare && cd ../client && npm ci && npm run prepare && cd ..
+  npm run prepare
```
jd2rogers2, line 4: look how simple now 😍

PR #200, opened 2022-08-20
152 asset dtos should not include user

esteban-gs:
## Dev Summary

Removing the `user: User` props from the assets dto and implementing authed routes to
- protect endpoint
- extract user from session

Only doing assets for now. I added another ticket for messages

Includes changes to tests and seeding

### changes not related to ticket, but to tooling
- added a jest config file to separate the unit test configs in `src` separate from the e2e/integration test configs in `test`. This change was needed as the `npm run test` was also picking up the tests in the `test` project

## Test Plan

See integration test for test cases and further steps
repro steps:
1. log in with insomnia or postman
2. save your cookie in your http client
3. make post request to `assets` with `json` body. If your http client is not sending the cookie from your log in request, grab it from the login response and add an http header for `Cookie` and for the value add the cookie the server returned on the log in request.

Expect to see an asset created without passing a user in the body of the request
Try it again without logging in or passing a cookie in the request and your should get a forbidden response


server/package.json
```diff
@@ -15,11 +15,11 @@
         "start:dev:db": "sudo docker-compose up",
         "start:prod": "node dist/main",
         "lint": "eslint src/**/*.ts",
-        "test": "jest",
+        "test": "jest --config ./src/jest.config.json",
```
esteban-gs, line 18: change to differentiate from e2e tests

server/package.json
```diff
@@ -15,11 +15,11 @@
...
         "test:watch": "jest --watch",
         "test:cov": "jest --coverage",
         "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-        "test:e2e": "jest --config ./test/jest-e2e.json",
+        "test:e2e": "jest --config ./test/jest-e2e.json --runInBand",
```
esteban-gs, line 22: for test tooling, run less asynchronously to avoid DB clashes

server/src/jest.config.json
```diff
@@ -0,0 +1,14 @@
...
+    }
+  },
+  "moduleFileExtensions": ["js", "json", "ts"],
+  "rootDir": ".",
+  "testEnvironment": "node",
+  "testRegex": ".spec.ts$",
+  "transform": {
+    "^.+\\.(t|j)s$": "ts-jest"
+  }
+}
```
esteban-gs, line 14: basically, being more explicit about what should be included as unit tests and not include other tests

server/test/assets/assets.e2e-spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  it('POST /assets -> when logged in -> asset should be created with logged in user', async () => {
+    const loginRes = await supertest
+      .agent(app.getHttpServer())
+      .post(`/auth/login`)
+      .send({ email: userSeed.email, password: userSeed.password })
+      .set('Content-Type', 'application/json')
+      .expect(201);
+    const cookie = loginRes.headers['set-cookie'];
```
esteban-gs, line 65: same approach to reproduce on `postman`

grab the cookie from response and add it to your subsequent `post` requests as an http `header` with key `Cookie` and paste the value

jd2rogers2 (approved):
- [x] plz remove root level package-lock.json
- [ ] inline comment(s)

couple small todos (above) first. but looks great! 💯 for the tests

server/src/auth/auth.controller.ts
```diff
@@ -1,4 +1,4 @@
-import { Controller, Post, Response, Request, UseGuards, Get } from '@nestjs/common';
+import { Controller, Post, Response, Request, UseGuards, Get, Logger } from '@nestjs/common';
```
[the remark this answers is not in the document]
esteban-gs: woops. I meant to remove this.

esteban-gs:
> * [x]  plz remove root level package-lock.json
> * [x]  inline comment(s)
> 
> couple small todos (above) first. but looks great! 100 for the tests

Thanks! I fixed the review issues

PR #218, opened 2022-10-03
Remove user dto

HartPM:
Dev Summary

Removing the user: User props from the messages dto and implementing authed routes to

protect endpoint
extract user from session

Includes changes to seeding. Did not update tests--please review to see if changes are needed for tests.


jd2rogers2 (approved):
LGTM!

- [ ] couple small in line comments
- [ ] make sure to run `npm run format` from server dir, then tests should pass and you should be able to push without the husky jobs commented
- [ ] ship it 🥳 🦜

Issue #208, opened 2022-09-10
user images step 1


jd2rogers2:
1. create user profile page or add following to user sign up page
2. create input for uploading photos
- needs to have validations that it's an image file and that the file size is not too big
3. on form submit sends photo to S3 bucket (create an AWS account and talk to @esteban-gs to get access to our project)
4. add `photo` property to image entity and DTOs
5. receives S3 API response with public image url that we then send to our BE as user's `photo`

https://github.com/aws/aws-sdk-js
https://blog.devgenius.io/upload-files-to-amazon-s3-from-a-react-frontend-fbd8f0b26f5
https://javascript.plainenglish.io/how-to-upload-files-to-aws-s3-in-react-591e533d615e

PR #220, opened 2022-10-07
Org sign up w/ image part 1a #208

jd2rogers2:
basic pieces of https://github.com/Nonprofit-Exchange-Hub/web-app/issues/208
does not include file size validation
does not include writing to AWS

adds image_url property to FE signup form data and validators, BE entity and req body validators

## Test Plan

repro steps:

1. http://localhost:3000/signup-nonprofit
2. fill out entire form including org profile image input (just use a url for now)
3. submit
4. see in server logs or via postgres `select id, name, ein, image_url from organizations order by id DESC limit 1;` that data, img made it to BE 

## Resources

- [opted not to allow heic uploads](https://itnext.io/tackling-iphone-or-ipad-images-support-in-browser-8e3e64e9aaa1)


client/src/components/Users/Auth/SignUpUserAndNonprofit/validation-rules.ts
```diff
@@ -51,7 +54,7 @@ export const calculateIsValid = (
   const valsAsStringArr = Object.values(vals);
   return (
     errVals.every((x) => x === undefined) &&
-    valsAsStringArr.length === 16 &&
+    valsAsStringArr.length === 17 &&
```
esteban-gs, line 57: \(^-^)/

esteban-gs (approved):
Nice! Approved ✔

Issue #207, opened 2022-09-10
Carousel component on homepage to use new Card component


jd2rogers2:
assuming both [card/theme PR](https://github.com/Nonprofit-Exchange-Hub/web-app/pull/201) and [carousel](https://github.com/Nonprofit-Exchange-Hub/web-app/issues/182) work have completed

the carousel work was done using placeholder cards and should now use the custom `Card` component we have
these cards should also show real data coming from the BE

PR #233, opened 2022-10-26
use cards in home page carousel #207

jd2rogers2:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


server/src/database/seeding/seed-data.ts
```diff
@@ -127,6 +127,46 @@ export function seedAssets(): CreateAssetDto[] {
       quantity: 12,
       imgUrls: [assetImg],
     },
+    {
+      title: 'Thirteen',
+      description: 'lucky lucky lucky',
+      type: AssetType.REQUEST,
+      condition: Condition.NONE,
+      quantity: 13,
+      imgUrls: [assetImg],
```
esteban-gs, line 136: Nice job! looks great. Could you change the image link to another public small image like any image from https://unsplash.com/s/photos/small-size-image? I know this isn't part of the ticket, but the link for `imageUrls` has been broken for some time

esteban-gs (approved):
@jessezenon , @jd2rogers2 good job! It looks really good

Except for that small change, it LGTM

server/src/database/seeding/seed-data.ts
jd2rogers2: @esteban-gs jesse and i both see that image on the FE so maybe you need to clear your browser cache??
otherwise, what do you mean by the img being broken?

esteban-gs: I think this may be related to how we load the image, then. Out of all the images in the app, my ad blocker stops this one image specifically from loading. But this shouldn't keep this PR from being merged.

skyfox93:
one issue with these needs cards is,  in the designs, it has the name and avatar of the organization that posted each asset. but we don't have this info. assets don't have an org_id. 
 we can add a query that takes a list of users and returns corresponding orgs, but there can be potential more than one org for each user. 
 or we can add an org_id to the asset

jd2rogers2 (commented):
1. 
> one issue with these needs cards is, in the designs, it has the name and avatar of the organization that posted each asset. but we don't have this info. assets don't have an org_id. we can add a query that takes a list of users and returns corresponding orgs, but there can be potential more than one org for each user. or we can add an org_id to the asset

you're right! we'll have to fix that in the long run. but for now we have it so that every org creates a new user, so it's a 1to1 relationship. so just get the first org from the user for now

2. could you fill out the PR description a bit more plz?

skyfox93:
I merged most of this PR in with mine already,  except a few server side changes,  so we can just accept all incoming on those client side conflicts.

skyfox93:
closing this PR since it looks like all relevant changes were already merged

Issue #212, opened 2022-09-10
asset images


jd2rogers2:
follows same process as https://github.com/Nonprofit-Exchange-Hub/web-app/issues/210
except asset resource needs to be able to have multiple images on it
- [ ] should be in different S3 bucket/sub-folder than the user and asset images
- [ ] do not allow duplicate file names

PR #235, opened 2022-11-01
Asset images #212

HartPM:
#212  

Dev Summary
Added a feature to enable the user to upload image urls when filling the Goods Needed form. Includes url validation and uniqueness of the url.

An array containing 1 empty string is the default state from which this portion of the form is rendered. For each element in the array, a text input field is rendered. Clicking a button to add another photo url appends 1 additional empty string to the array, thereby rendering an additional input field. The user should add one image url to each field. If the user enters a string that is not a valid url, an error message is displayed by <FormHelperText>.

Yup validation schemas were created in a separate file, NeedFormGoods-validation.ts. urlSchema is used to check each image url is a valid url. I included a custom method to check if each url is unique that runs on the validationSchema when the form is submitted, but this needs to have additional work done to display an error to the user if we want to keep this check.

Test Plan
1. Entered many valid and invalid urls into the input fields
2. Attempted entering duplicate urls
3. Confirmed the form submission successfully posted images to the backend

repro steps:

1. http://localhost:3000/need/goods
2. Scroll to very bottom of the form to see input fields below "Or link photos below"
3. Add first image url to input field labeled "Photo 1"
4. Click button "Click here to add another photo"
5. Repeat the process for the desired number of images to upload.
6. Fill out the rest of the form and submit it.
expected: http://localhost:3001/api/assets -- latest object includes "imgUrls" property with an array of the urls entered on the frontend.

Screen Shot 2022-10-31 at 9 08 59 PM

Screen Shot 2022-10-31 at 9 11 39 PM


client/src/views/NeedFormGoods.tsx
```diff
@@ -83,6 +87,47 @@ function NeedForm(): JSX.Element {
...
 
+  const isValidUrl = (urlString: string) => {
+    let urlPattern = new RegExp(
+      '^(https?:\\/\\/)?' +
+        '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' +
+        '((\\d{1,3}\\.){3}\\d{1,3}))' +
+        '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' +
+        '(\\?[;&a-z\\d%_.~+=-]*)?' +
+        '(\\#[-a-z\\d_]*)?$',
+      'i',
```
esteban-gs, line 98: impressive attempt at validating the url!

HartPM:
I replaced the original validations I wrote with Yup validations. I couldn't figure out a way to validate the uniqueness of urls until form submit. It looks like error messaging display is still needed within this component-- see line 170. Also noticed the category drop down on the form doesn't work.

esteban-gs (approved):
I think you could extract the url list form to another child component, but given that this is a temporary fix, approved!

PR #238, opened 2022-11-03
fix double api in auth path

skyfox93:
## Dev Summary

Fixes double "/api/api" in authorization check 
Screen Shot 2022-11-02 at 5 53 02 PM

## Test Plan

repro steps:

1. localhost:3000 
2. login
3. navigate to homepage
 
expected: 
user should still be logged in

## Resources

### DIFF
--- a/client/src/providers/UserProvider.tsx
+++ b/client/src/providers/UserProvider.tsx
@@ -21,7 +21,7 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   const [user, setUser] = React.useState<User | null>(null);
 
   async function fetchUser(): Promise<User | null> {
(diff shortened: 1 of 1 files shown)


client/src/providers/UserProvider.tsx
```diff
@@ -21,7 +21,7 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   const [user, setUser] = React.useState<User | null>(null);
 
   async function fetchUser(): Promise<User | null> {
-    const res = await fetch(`${APP_API_BASE_URL}/api/auth/session`, {
+    const res = await fetch(`${APP_API_BASE_URL}/auth/session`, {
```
esteban-gs, line 24: Good catch!

esteban-gs (approved):

Issue #209, opened 2022-09-10
user images step 2


jd2rogers2:
get user's image url (string) from BE and displays (via `<img` tag) in some places on the FE
- logged in user img in header
- user profile page?

PR #240, opened 2022-11-04
#209 [WIP] user images step 2

esteban-gs:
# Important
 **blocked by: https://github.com/Nonprofit-Exchange-Hub/web-app/pull/264**
## Dev Summary
# Not ready for review. This work is blocked by 
### Changes
- [x] Refactored `AuthModule` and `UsersModule` into a single account manager module since these play a similar role and are usually imported together to get any auth functionality, it make sense to keep them together. It makes testing easier: https://blog.logrocket.com/avoid-circular-dependencies-nestjs/
- [x] Added a top level `FileService` that implements Azure blob storage for persisting files. The caveat with this is that there is still work to be done to replace all files, since files are saved with an extension, there could be x number of extension possible files to check.
- [x] Added an avatar component to the UI that renders the file if the user has a profile image
- [x] Added profile image to the global `User` type in react  

### not implemented
implementation for local dev like saving to local dev environment, maybe later

## Test Plan

repro steps:

1. Start the app
2. using an http client like insomnia, log into the route: `http://localhost:3001/api/auth/login`, when successful this will save the cookie
3. Make a post request to `http://localhost:3001/api/auth/profile/{userid}` with multipart form body:
  - include a form property `profile_image_url` and use insomnia or postman to include an image file in the form request (smaller than 1MB)

Before testing, contact me to get the Azure api key
Expect to see the image as an avatar for the user when you log in as the same user in react. The file will be saved a public Azure blob (inspect the image url)

## Resources


client/src/components/Header.tsx
```diff
@@ -232,7 +232,10 @@ function Header() {
                 onClick={handleClick}
                 size="large"
               >
-                <AccountCircleIcon />
+                <UserAvatar
+                  userFirstName={user.firstName}
+                  profileImageUrl={user.profile_image_url ?? ''}
+                />
```
esteban-gs, line 238: display logic is pushed to child component

client/src/components/Users/Auth/ForgotPassword.tsx
```diff
@@ -92,7 +92,7 @@ function ForgotPassword() {
     evt.preventDefault();
     setIsLoading(true);
 
-    await fetch(`${APP_API_BASE_URL}/api/users/reset_password`, {
+    await fetch(`${APP_API_BASE_URL}/api/auth/reset_password`, {
```
esteban-gs, line 95: all the auth things are under the `auth` route now

client/src/views/SignupCitizen.tsx
```diff
@@ -91,7 +91,7 @@ function SignupCitizen() {
     setIsLoading(true);
     // Backend doesn't need accept_terms. If a user is signed up they have agreed to the terms
     delete formData.accept_terms;
-    const res = await fetch(`${APP_API_BASE_URL}/users`, {
+    const res = await fetch(`${APP_API_BASE_URL}/auth/register`, {
```
esteban-gs, line 94: all auth routes are in the account manager module now

server/sample.env
```diff
@@ -0,0 +1,22 @@
...
+
+# e2e used when running e2e tests
+E2E_DATABASE_DB=e2e_db
+
+# sockets gateway cors
+SOCKET_CORS_ORIGIN="http://localhost:3000"
+
+# Azure blob storage
+AZURE_BLOB_CONNECTION_STR=some-string
+AZURE_BLOB_CONTAINER=images
```
esteban-gs, line 22: Probably best to persist a sample `.env` file at this point

server/src/account-manager/guards/cookie-auth.guard.ts
```diff
@@ -26,6 +26,7 @@ export class CookieAuthGuard extends AuthGuard() {
     try {
       user = await this.jwtService.verify(jwt, { secret: jwtConstants.secret });
     } catch (_e) {
+      Logger.error(_e, CookieAuthGuard.name);
```
esteban-gs, line 29: This is useful for debugging

Issue #223, opened 2022-10-19
clean up homepage


jd2rogers2:
screen shots TBD

moving the search bar to home component, and that should be above hero content
remove old hero content
new hero content (already exists) should already land under search bar, but if not align it there

PR #243, opened 2022-11-06
update Home view as requested in ticket 223

jinook929:
## Dev Summary
- Relocate the search box at Home view [here](https://github.com/Nonprofit-Exchange-Hub/web-app/issues/223)
- Update devDependency of "@types/react" to "^18.0.9" for client

## Test Plan

repro steps:

1. localhost:3000 
2. see if the search box has been relocated as in [Figma](https://www.figma.com/file/KE2Sw6ob07agN25RmeNuFg/GiveCycle-wireframes?node-id=6351%3A41839)
image

### DIFF
--- a/client/src/views/Home.tsx
+++ b/client/src/views/Home.tsx
@@ -15,6 +15,7 @@ import BannerSection from '../components/BannerSection';
 import routes from '../routes';
 
 import type { Asset } from '../types';
+import { APP_API_BASE_URL } from '../configs';
 
 const carouselSteps = [
   [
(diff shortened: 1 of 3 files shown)


client/src/views/Home.tsx
```diff
@@ -206,28 +204,18 @@ function HeaderContentRight(): JSX.Element {
...
-
-    const assetsApiRequest = new URL(ASSETS_API_URL);
-    assetsApiRequest.searchParams.append('type', 'request');
-    assetsApiRequest.searchParams.append('limit', '3');
-    assetsApiRequest.searchParams.append('offset', '0');
-    fetch(assetsApiRequest.href)
+    fetch('http://localhost:3001/api/assets?type=request&limit=3&offset=0')
       .then((resp) => resp.json())
       .then((data: Asset[]) => {
         setRequests(data);
```
esteban-gs, line 221: Can you use the same url that was in the original code? That will ensure that it works in staging as well. Also, instead of passing static strings as the query params, can you use the existing example?

esteban-gs (approved):
Looks great! Approved with some small changes
- [x] One small change in comments
- [ ] and can you also make the input field slightly larger radius and add a little more shadow to resemble the mocks more closely?
- [ ] If you can, use the theme to make this change

client/src/views/Home.tsx
jinook929: @esteban-gs pushed the update as requested~

PR #256, opened 2022-11-15
226 org sign up phone number syntax UI bug

esteban-gs:
## Dev Summary
Implement input masking for
- Org sign up phone
- Org sign up state 

I had to install the latest version of `react-input-mask` library to get this to work with our other libraries, `react-hooks-form` and `MUI`. There is an error triggered by the `React.Strict` config - it's a known issue.

## Test Plan
repro steps:

1. Fill out the Org sign up form
2. get to the steps where you fill out the phone and state
3. Expect to see the following errors for wrong phone and state format including the error descriptions
image

4. Expect the errors to not display when you enter valid data:
image

## Resources

- [why use the alpha version of `react-input-mask`](https://stackoverflow.com/a/72114444/13862254)


client/src/components/Users/Auth/SignUpUserAndNonprofit/validation-rules.ts
```diff
@@ -2,11 +2,12 @@ import { DeepRequired, FieldErrorsImpl } from 'react-hook-form';
 import { FormData } from './FormData';
 import * as Yup from 'yup';
 import { classifications } from './Classifications';
+import { US_STATE_ABBREVIATIONS } from '../../../../configs';
 
 export const validationSchema = Yup.object().shape({
   doing_business_as: Yup.string().required('Required'),
   city: Yup.string().required('Required'),
-  state: Yup.string().required('Required').min(2).max(2),
+  state: Yup.string().required('Required').oneOf(US_STATE_ABBREVIATIONS),
```
jd2rogers2, line 10: looks like `.oneOf` can take an error message as the 2nd arg as well
https://github.com/jquense/yup#schemaoneofarrayofvalues-arrayany-message-string--function-schema-alias-equals
i'd vote to pass in a string like `"must be a valid US state abbreviation"` or something

jd2rogers2 (approved):

PR #258, opened 2022-11-18
Add carousels

skyfox93:
## Dev Summary

[Mocks](link)

## Test Plan

 - Introduces a Responsive Carousel component , including:
  - Adjusts the number of cards when the page width is too narrow, instead of "squashing" the content of the cards.
  - Carousel now holds its own state
  - content is determined by passing in a renderItem function and a fetch method function
  
  Card changes
  -  Limits the height of the description in a card so that the card does not get too long
  -  improves formatting for "posted by"
  - may revert some changes due to confusion over which design to use
  
 
Screen Shot 2022-11-17 at 11 03 13 PM

  
Screen Shot 2022-12-22 at 11 16 04 AM
Screen Shot 2022-12-22 at 11 16 12 AM

  
  
  
  -Changes to the home Component
  - moves fetch functions into a module called "fetchActions.js" for reusibility
  - removes state from the home component as it now belongs to the carousel component
  
  

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


client/src/components/ResponsiveCarousel.tsx
```diff
@@ -0,0 +1,157 @@
+import * as React from 'react';
```
jd2rogers2, line 1: tip for next time: wait until after review approval to change file names
that way reviewer can see changes you've made within the file

jd2rogers2 (approved):
everything looks great! sorry for the slowness

PR #259, opened 2022-11-29
password reset flow

jd2rogers2:
## Test Plan

repro steps:

get a sendgrid api key add to your `.env` as `SENDGRID_API_KEY`
add a `JWT_SECRET` to your `.env` file too (any value really)

1. http://localhost:3000/forgot-password
2. enter email, submit
expected: receive email in your inbox


server/src/auth/auth.module.ts
```diff
@@ -17,7 +16,7 @@ import { WSCookieStrategy } from './strategies/ws-cookie..strategy';
     UsersModule,
     PassportModule,
     JwtModule.register({
-      secret: jwtConstants.secret,
+      secret: process.env.JWT_SECRET,
```
esteban-gs, line 19: ✔

server/src/sendgrid/sendgrid.service.ts
```diff
@@ -0,0 +1,15 @@
...
+
+@Injectable()
+export class SendgridService {
+  constructor() {
+    SendGrid.setApiKey(process.env.SENDGRID_API_KEY);
+  }
+
+  async send(mail: SendGrid.MailDataRequired) {
+    const transport = await SendGrid.send(mail);
+    console.log(`E-Mail sent to ${mail.to}`);
```
esteban-gs, line 12: Thanks for logging this event. Can you use the NestJs `Logger` class, it will print other metadata. 

```ts
Logger.log(`E-Mail sent to ${mail.to}`, SendgridService.name);
```

server/src/users/users.controller.ts
```diff
@@ -35,10 +41,32 @@ export class UsersController {
     @Response({ passthrough: true }) response: ResponseT,
   ): Promise<void> {
     try {
-      // const user = await this.usersService.findByEmail(req.body.email);
-      // TODO send email the user
+      const user = await this.usersService.findByEmail(req.body.email);
+      if (!user) {
+        throw new Error();
```
esteban-gs, line 46: This should be
```suggestion
        return new NotFoundException(`foo bar`);
```

esteban-gs (approved):
Nice! Approved once the comments are resolved 🚀

server/src/users/users.controller.ts
jd2rogers2: left a comment inline for clarification
this is in a `try catch` so it doesn't really matter what error we throw here
and the catch responds 200 so that all those hackerz can't find out which emails are active or not on our service

PR #264, opened 2023-01-15
Feature/account manager module

esteban-gs:
## Dev Summary

The proposal of this PR is to combine the `auth` and `user` module into a new `account-manager` module. 

My general approach:
- add the new module and use a temporary V2 suffix with the new version of those services
- add deprecation errors to the old code pointing to the new module
- for change management, it will be easier to add a temporary prefixed module (instead of completely removing them) while we move things from old to new modules. Once we have dealt with all merge conflicts, it should be easy to remove the old modules all together and remove the temporary suffix.

App should behave as it was previously, except, the following endpoints have been updated:

- `/users` -> `/auth/users` (this PR only changes the backend)

## Test Plan

repro steps:

1. I am verifying that my changes have not broken the app thanks to the e2e tests. Very detailed tests pass there

## Other
- Before we merge this ticket, we should create a corresponding card to update the front end routes
- After this is merged into main and other developers have had a chance to integrate this into their work, we should add a ticket for:
 - removing the suffixes
 - removing old modules


server/src/acccount-manager/strategies/ws-cookiev2..strategy.ts
```diff
@@ -0,0 +1,6 @@
+import { Strategy } from 'passport-local';
```
jd2rogers2, line 1: filename has 2 periods in it

server/src/auth/auth.controller.ts
```diff
@@ -1,27 +1,28 @@
...
   constructor(private authService: AuthService) {}
 
   @Post('login')
-  @UseGuards(LoginAuthGuard)
   async login(
     @Request() request: AuthedRequest,
     @Response({ passthrough: true }) response: ResponseT,
   ): Promise<void> {
     const { user } = request;
+    throw new BadRequestException('Deprecated resource. Use the new account manager module');
```
jd2rogers2, line 25: throwing errors in here is the same as removing imo
the impact to the frontend is the same
i vote we either keep these as they were or delete now

jd2rogers2: and same goes for all other endpoints that `throw`

jd2rogers2 (approved):
- [ ] still need to handle seeding data with the new endpoints
- [ ] some inline comments
but otherwise looks good!

server/src/auth/auth.controller.ts
esteban-gs: Mainly, I left this here to make other engineer's conflict resolution easier.

server/src/acccount-manager/strategies/ws-cookiev2..strategy.ts
esteban-gs: Good catch!

PR #267, opened 2023-01-20
ryan-werner-branch

rmwerner18:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


client/src/components/Header.tsx
```diff
@@ -69,7 +69,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 }));
 
 function Header() {
-  const LOGOUT_URL = `${APP_API_BASE_URL}/logout`;
+  const LOGOUT_URL = `${APP_API_BASE_URL}/auth/logout`;
```
esteban-gs, line 72: Nice! Good catch

client/src/components/PrivateRoute/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { useContext } from 'react';
+import { Redirect } from 'react-router-dom';
+import { UserContext } from '../../providers';
+import routes from '../../routes';
+
+const PrivateRoute = ({ children }) => {
```
esteban-gs, line 6:
```suggestion

type Props = {
  children: JSX.Element;
};

const PrivateRoute = ({ children }: Props) => {
```

client/src/views/Main.tsx
```diff
@@ -47,42 +49,49 @@ function Main() {
...
 
-        {/* action: needs */}
-        <Route exact path={routes.NeedForm.path} component={routes.NeedForm.component} />
-        <Route exact path={routes.NeedFormGoods.path} component={routes.NeedFormGoods.component} />
-        <Route
-          exact
-          path={routes.NeedFormVolunteers.path}
-          component={routes.NeedFormVolunteers.component}
-        />
+        <PrivateRoute>
```
esteban-gs, line 62: You need to wrap the `<PrivateRoute>` children in a React fragment
````tsx
<>
  <Route...../>
</>```

esteban-gs (changes requested):
Good job! We get a lot of functionality from this component. I think we can make use of this component to expand the functionality. 

- Move the private route component in a new directory `src/routes/PrivateRouteWrapper/index.tsx` 

The next step is to add a roles property (or something that may be more appropriate) to the routes in `src/routes`. Then, in your `PrivateRoutes` component, you can check if the children passed are in the routes object, if so, we can check if the user has the roles for this route.

esteban-gs (approved):

PR #269, opened 2023-01-26
HTTP response logger middleware

esteban-gs:
## Dev Summary

This PR adds an HTTP response middle ware. It's part of the ticket. 

#### Not implemented
Log levels for different environments

## Test Plan

repro steps:

1.  start the server
2. Make any http request to an `/api` endpoint
3. Expect to see the http response logged
4. Expect to NOT see the static assets http requests logged. In other words, when you navigate through the front end, only the requests to the `api/` route is logged 

## Resources

n/a


server/src/app.module.ts
```diff
@@ -25,6 +26,7 @@ import { SendgridService } from './sendgrid/sendgrid.service';
     TypeOrmModule.forRootAsync({ useClass: DatabaseConnectionService }),
     ServeStaticModule.forRoot({
       rootPath: join(__dirname, '/../../client', 'build'),
+      exclude: ['/api*'],
```
esteban-gs, line 29: this tells the server not to serve any static assets from `api/`

server/src/main.ts
```diff
@@ -11,18 +10,14 @@ dotenv.config({ path: __dirname + '/.env' });
...
   app.enableCors({
     credentials: true,
     // TODO get env related base url
     origin: 'http://localhost:3000',
   });
   app.useGlobalPipes(new ValidationPipe({ transform: true }));
-  app.useStaticAssets(resolve('./src/public'));
-  app.setBaseViewsDir(resolve('./src/views'));
-  app.setViewEngine('hbs');
   app.use(cookieParser('secret_placeholder'));
```
esteban-gs, line 20: Removing these because we don't use them. These are only needed if we were serving MVC views

jd2rogers2 (approved):

PR #271, opened 2023-01-27
Use LIKE in assets search

skyfox93:
Use LIKE query for title instead of title=
## Test Plan

- Go to the asset search bar. set it to "needs" and search "eggs" 
there will be a result, "one dozen eggs"

### DIFF
--- a/client/src/views/SearchResults.tsx
+++ b/client/src/views/SearchResults.tsx
@@ -104,7 +104,7 @@ function SearchResults(): JSX.Element {
     } else if (querySearchCategory === 'Offers' || querySearchCategory === 'Needs') {
       fetch(
         `${APP_API_BASE_URL}/assets?type=${
-          querySearchCategory === 'Needs' ? 'donation' : 'request'
+          querySearchCategory === 'Needs' ? 'request' : 'donation'
         }${querySearchText ? `&title=${querySearchText}` : ''}`,
       )
         .then((resp) => resp.json())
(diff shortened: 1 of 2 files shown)


server/src/assets/assets.service.ts
```diff
@@ -21,11 +22,12 @@ export class AssetsService {
   }
 
   async getAssets(getAssetsDto: GetAssetsDto): Promise<Asset[]> {
-    const { limit, offset, ...rest } = getAssetsDto;
+    const { limit, offset, title, ...rest } = getAssetsDto;
+    const whereParams = title ? { title: Like(`%${title}%`), ...rest } : { ...rest };
```
esteban-gs, line 26: Thanks for looking at this. Getting the search functionality to work is fantastic. 

One thing about `Like` is that it is only a temporary work around to doing more performant queries like full-text search.

client/src/views/SearchResults.tsx
```diff
@@ -104,7 +104,7 @@ function SearchResults(): JSX.Element {
     } else if (querySearchCategory === 'Offers' || querySearchCategory === 'Needs') {
       fetch(
         `${APP_API_BASE_URL}/assets?type=${
-          querySearchCategory === 'Needs' ? 'donation' : 'request'
+          querySearchCategory === 'Needs' ? 'request' : 'donation'
```
esteban-gs, line 107: The preferred url and query param creation is with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

esteban-gs (changes requested):
Hi @skyfox93 , Looks good! Thanks for working on this. Just one small change

client/src/views/SearchResults.tsx
jd2rogers2: 😮‍💨🌴 i was wondering why the search bar wasn't returning the results i expected.....
great catch

PR #272, opened 2023-01-30
add progress steps to SignupCitizen

rmwerner18:
No validations yet before switching to next page in form. Also, a lot of code copied from SignUpUserAndNonprofit.tsx, can probably refactor.


esteban-gs (changes requested):
Hi @rmwerner18 , great work! We can take a step further and add the fields that are visible in the mocks like `city` `state` `zip` and `photo` and add them just as a visual element, so not wired up to the working form. Except the interest step

client/src/views/SignupCitizen.tsx
```diff
@@ -91,6 +160,7 @@ function SignupCitizen() {
 
   const handleSubmit = async (evt: React.FormEvent) => {
     evt.preventDefault();
+    handleNext();
```
jd2rogers2, line 163: let's put this in the else below
and remove that history.push
that way we see the success page and aren't automatically move to the home screen

client/src/views/SignupCitizen.tsx
```diff
@@ -110,132 +180,361 @@ function SignupCitizen() {
...
+                  className={header}
+                  variant="h4"
+                  fontSize="58px"
+                  component="h1"
+                  align="left"
+                >
+                  Sign up Complete!
+                </Typography>
+                <Typography className={label} sx={{ fontWeight: 'bold' }}>
+                  Your name {/* USER'S NAME GOES HERE */}
```
jd2rogers2, line 463: you can get the name and email of the user from the response
per this endpoint https://github.com/Nonprofit-Exchange-Hub/web-app/blob/f2c3e01df8c171c7ff3d0d57674467f1af073cc5/server/src/acccount-manager/account-manager.controller.ts#L47

client/src/views/SignupCitizen.tsx
```diff
@@ -91,6 +160,7 @@ function SignupCitizen() {
 
   const handleSubmit = async (evt: React.FormEvent) => {
     evt.preventDefault();
+    handleNext();
     setIsLoading(true);
     // Backend doesn't need accept_terms. If a user is signed up they have agreed to the terms
     delete formData.accept_terms;
```
jd2rogers2, line 166: while we're here let's change the fetch line to use @esteban-gs 's new endpoints 🥳 
so in [this new account-manager controller](https://github.com/Nonprofit-Exchange-Hub/web-app/blob/f2c3e01df8c171c7ff3d0d57674467f1af073cc5/server/src/acccount-manager/account-manager.controller.ts) it looks like we should be hitting `/users/register`

jd2rogers2 (approved):
before merging: couple of small tweaks that i've made inline comments for
then it's good to go. nice work!!

garimacha:
[User Sign Up](https://trello.com/c/1iY3O8aR/3-user-sign-up)

client/src/views/SignupCitizen.tsx
jd2rogers2: sorry, i think it's `/auth/register` actually
cc: @esteban-gs to confirm

esteban-gs: YEs!

esteban-gs (approved):

PR #275, opened 2023-02-01
168 assetspage

alexislcstone:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


package-lock.json
```diff
@@ -0,0 +1,64 @@
+{
```
esteban-gs, line 1: This entire file along with the `packag.json` can be removed

server/src/permissions/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+  admin = 'ADMIN',
+  owner = 'OWNER',
+  revoked = 'REVOKED',
+}
+
+export enum ApprovalStatus {
+  approved = 'APPROVED',
+  pending = 'PENDING',
+  denied = 'DENIED',
+}
```
esteban-gs, line 11: There are a lot of changes showing up that aren't part of your PR. Let's pair to rebase your branch to `main`

client/src/components/Card/NeedCard.tsx
```diff
@@ -38,16 +38,18 @@ export default function NeedCard(props: Props) {
...
     >
-      <CardContent>
-        <Box sx={{ marginTop: theme.spacing(1), padding: '0.5em' }}>
-          <Tooltip title={props.title}>
-            <Typography variant="h1" sx={ellipsesStyle}>
-              {props.title}
-            </Typography>
-          </Tooltip>
+      <CardContent sx={{ margin: '5% 5% 0% 5%' }}>
+        <Box sx={{ display: 'flex', justifyContent: 'space-between ' }}>
```
esteban-gs, line 45:
```suggestion
        <Box sx={{ display: 'flex', justifyContent: 'space-between' }}>
```

esteban-gs (changes requested):
Nice job! It looks good. A couple things you should revise:
- Remove comments with unused except for the comments with `TODO:`
- All the text in the card seems to be underlined - it shouldn't all be

client/src/components/Card/NeedCard.tsx
```diff
@@ -7,47 +7,34 @@ import Typography from '@mui/material/Typography';
...
-      sx={props.sx || {}}
-      org={props.poster.firstName}
-    >
-      <CardContent>
-        <Box sx={{ marginTop: theme.spacing(1), padding: '0.5em' }}>
-          <Tooltip title={props.title}>
-            <Typography variant="h1" sx={ellipsesStyle}>
-              {props.title}
-            </Typography>
-          </Tooltip>
```
skyfox93, line 50: I made these changes earlier, they should not be reverted. The title should not be hardcoded as "Blazers", and the title should take up only 1 line, with a tooltip to indicate the full title.  The terms "apparel" and "long term"  shouldn't be hardcoded either.  We can wrap these in some conditional logic to show {asset.category} and {asset.urgency}  if present, but I felt having all this placeholder text makes it look like we support features we don't support, and makes it harder to determine what work still needs to be done. That's why I left those for a separate PR adding those features in.  If you want placeholders though, can we do something like "{category}" "{urgency}", so thats it clear we need to work on those?

skyfox93 (changes requested):
See comment on reverted changes to AssetCard

client/src/components/Card/NeedCard.tsx
alexislcstone: All of this make sense! Thanks for changing it and the explanation!

garimacha: @alexislcstone  are all the comments addressed? Is this ready to be closed and merge?

EtoKruto: @garimacha I've taken over this feature branch, addressed comments and finalized where Alexis has paused her work.

garimacha: Thank you @EtoKruto is there any other ticket you are also working on? I need details to update my planning document. Thanks.

EtoKruto:
@esteban-gs or @jd2rogers2 please review the package-lock.json I'm not sure how to resolve this safely so I'd rather revert to your guidance on this. This issue should be completed unless I'm missing minor details with regard to translating Figma to code

jd2rogers2 (approved):

PR #277, opened 2023-02-06
fix private route bug

rmwerner18:
Fixed bug that was happening on reload in private routes. The page used to redirect to Login even if user was logged in, because redirecting was happening before user object was loaded. I added an 'isLoading' attribute to the UserContext to prevent this. UserContext now looks like {user, setUser, isLoading} instead of [user, setUser].

### DIFF
--- a/client/src/providers/UserProvider.tsx
+++ b/client/src/providers/UserProvider.tsx
@@ -41,19 +41,17 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   }
 
   async function setUserTimeout(
-    user: User | null,
+    currentUser: User | null,
     shouldFetch = false,
     shouldStartTimer = false,
   ): Promise<void> {
-    let newUser: User | null = user;
+    let newUser: User | null = currentUser;
     if (shouldFetch) {
-      setIsLoading(true);
       newUser = await fetchUser();
     }
 
     setUser(newUser);
-
-    if (isLoading) setIsLoading(false);
+    setIsLoading(false);
 
     if (shouldStartTimer && newUser) {
       setTimeout(() => {
--- a/client/src/providers/UserProvider.tsx
+++ b/client/src/providers/UserProvider.tsx
@@ -8,17 +8,23 @@ type SetUser = (
   shouldFetch?: boolean,
   shouldStartTimer?: boolean,
 ) => Promise<void>;
-type UserContextT = [User | null, SetUser];
+type UserContextT = {
+  user: User | null;
+  setUser: SetUser;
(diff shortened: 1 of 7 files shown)


client/src/providers/UserProvider.tsx
```diff
@@ -42,10 +47,14 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   ): Promise<void> {
     let newUser: User | null = user;
     if (shouldFetch) {
+      setIsLoading(true);
       newUser = await fetchUser();
     }
+
     setUser(newUser);
 
+    if (isLoading) setIsLoading(false);
```
jd2rogers2, line 56: so react states actually work async under the hood, the do some batching and stuff and basically we can't rely on the `isLoading` state to be `true` here
but i do think we could just set it to false without the condition as regardless of whether we even fetched, it should still be false at this point

client/src/providers/UserProvider.tsx
```diff
@@ -42,10 +47,14 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   ): Promise<void> {
     let newUser: User | null = user;
     if (shouldFetch) {
+      setIsLoading(true);
```
jd2rogers2, line 50: feel free to test this theory by changing the setTimeout timer to like 10 seconds or something...

the way this works is it automatically makes a request to get a new token right before the token is about to expire (1hr)
at that time i think this would cause our page to show the "loading" text you've added in `PrivateRouteWrapper/index.tsx`
imo the user won't want to see the loading text flash in the middle of their session so i think we should only set this isLoading on our initial fetch
i think wrapping this setter in an `if newUser === null` will work

client/src/providers/UserProvider.tsx
```diff
@@ -42,10 +47,14 @@ export function UserProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   ): Promise<void> {
```
jd2rogers2, line 47: while you're here can you change this terrible code (written by me)
the `user` arg here clashes with the `user` state above
maybe name this one `currentUser`??

jd2rogers2 (approved):
couple small changes to make but then this is ready to 🚢 
nice work!!

client/src/providers/UserProvider.tsx
rmwerner18: Nice catch, I figured out that this line doesn't actually need to be there at all since the loading state is defaulted to true

jd2rogers2: even better!

PR #281, opened 2023-02-12
email verification

jd2rogers2:
## Dev Summary

finally! sendgrid account with npc email and hooked it up to sign up endpoint

## Test Plan

repro steps:

1. get sendgrid api key from james
2. add FE_DOMAIN=http://localhost:3000 to the `.env` file
3. http://localhost:3000/signup
4. sign up with real email
5. try to sign in and fail
6. go to your email and verify
7. now sign in
8. success

## Resources
- https://docs.nestjs.com/fundamentals/testing#auto-mocking


esteban-gs (approved):
@jd2rogers2 , Very nice!! this is really neat to see in action.

It seems like the e2e test is failing because it expects the sendgrid service to actually send the email. You could  get past this by mocking the sendgrid service in that e2e test.

Long term, we should look at implementing an `EmailService` that abstracts the `Sendgrid` service. We could set up [mailtrap](https://mailtrap.io/email-sandbox/) for local dev and sendgrid for production. This will help with testing in the ci pipeline as well.

For now, I think it's best to just mock the sendservice so the test can pass

- [ ] see if you can make that test pass by mocking the sendgrid service

PR #282, opened 2023-02-15
Faqs Section

brodylosh:
## Dev Summary

[Figma Mocks](https://www.figma.com/file/K7DT5jYcLKRc9epmNrO0N7/UX---Dev-Handoff?node-id=273%3A3675&t=gUF3Yj52JdNpXPNS-0)

Updated the FAQs section of the homepage to reflect the new aesthetic of the Figma mocks. No functionality impacted, primarily just styling. This should provide users with a much more appealing UI. I noticed that others had started to create separate React components for different sections of the homepage, so I followed suit, to help compartmentalize the code. Additionally, I created an Icons folder within Components to store a newly created AccordionIcon (see resource section), that is used to expand the questions within FAQs. I also created the structure for the FAQs image using a placeholder image, to be replaced by the real image, as soon as it is made available.

## Test Plan

No functionality impacted, only styling. Can check that Accordion Icons are working properly.

Repro steps:

1. localhost:3000
2. Scroll to FAQs (above headers) and click on each question
3. Ensure that each question expands, and that the div does not overflow into the Footer (it should expand if necessary)

## Resources

[Stack Overflow - Accordion Icon](https://stackoverflow.com/questions/63691127/how-to-change-icon-when-accordion-summary-is-expanded)


jd2rogers2 (commented):
looks great!! 
1. can we replace the old placeholder image with one of the ones [from the drive](https://drive.google.com/drive/u/0/folders/17TyKrqROm9epjrv_cQvvQwNmniNInkq5) for now
i know none are the actual one we want but they're closer
2. a little bit of image styling
```
// on faqsImage, **remove everything else** and add:
    object-fit: none; // and try all the other object-fit values here too to find which is best
    width: 100%;
    height: 100%;
// on faqsVisual
    padding: 20px;
```

should be a quick approve after these changes, thanks a ton it's really looking good

jd2rogers2 (approved):

PR #283, opened 2023-02-18
use full certs

esteban-gs:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


jd2rogers2:
@esteban-gs what do these values do?

esteban-gs:
those are database connection ssl certificates. I'm looking at different cloud providers and needed to include those to test a deployment

PR #294, opened 2023-02-24
add cols to user and org

rmwerner18:
Made facebook, instagram, and twitter optional in orgs, and made bio optional in users. Also added this data to existing seed data.


esteban-gs (commented):
It looks like you may have to fix some tests to account for the new fields. We can probably pair.

esteban-gs (approved):
Nice!

PR #295, opened 2023-02-28
swagger like mick jagger

jd2rogers2:
## Dev Summary

we've got swagger docs now!!

Screen Shot 2023-02-27 at 5 25 11 PM

## Test Plan

repro steps:

1. http://localhost:3001/docs
test out the endpoints

## Resources

- [nest docs](https://docs.nestjs.com/openapi/introduction)

### DIFF
--- a/server/src/main.ts
+++ b/server/src/main.ts
@@ -16,8 +16,7 @@ async function bootstrap() {
   app.setGlobalPrefix('/api');
   app.enableCors({
     credentials: true,
-    // TODO get env related base url
-    origin: 'http://localhost:3000',
+    origin: process.env.FE_DOMAIN,
   });
   app.useGlobalPipes(new ValidationPipe({ transform: true }));
   app.use(cookieParser('secret_placeholder'));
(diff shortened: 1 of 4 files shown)


server/src/main.ts
```diff
@@ -3,21 +3,33 @@ import { NestFactory } from '@nestjs/core';
...
   });
   app.useGlobalPipes(new ValidationPipe({ transform: true }));
   app.use(cookieParser('secret_placeholder'));
+
+  const config = new DocumentBuilder()
+    .setTitle('Nonprofit Circle API')
+    .setDescription('Our API monolith')
+    .setVersion('0.0')
+    .addTag('nonprofit')
+    .build();
```
esteban-gs, line 29: We should also add the options to save an auth cookie. 
```suggestion
    .setVersion('0.0')
    .addTag('nonprofit')
    .addCookieAuth('NEH_is_cool')
    .build();
```

esteban-gs (approved):
This is super helpful! Glad to see that Nestjs can parse out the http endpoints. We still need to re-write some controller to make use of dtos so that they can be auto-documented by nest. One example of these controllers is the `auth/login` controller - it can't be understood by swagger.

We can probably review this in another iteration. Or if you want to go for it, we should try to re-write at least the account controllers.

PR #300, opened 2023-03-12
clean up search bar and share on home and search results pages

jd2rogers2:
## Dev Summary

the home page search bar styling was broken and it wasn't using the same component as the search results page
the good search bar itself also needed some tweaks
removed the tooltip (inherently bad accessibility)
also fixed the 'view more' buttons on the homepage

## Test Plan

repro steps:

1. localhost:3000
2. use search bar a bunch
3. use search bar a bunch from /SearchResults page

expect it to work intuitively for you

### DIFF
--- a/client/src/components/Search.tsx
+++ b/client/src/components/Search.tsx
@@ -40,7 +40,7 @@ function Search() {
   const [searchText, setSearchText] = React.useState<string>((querySearchText as string) || '');
 
   const handleSearch = () => {
-    history.push(`/SearchResults?search=${searchText}&category=${searchCategory}`);
+    history.push(`/search-results?search=${searchText}&category=${searchCategory}`);
   };
 
   return (
--- a/client/src/components/Search.tsx
+++ b/client/src/components/Search.tsx
@@ -1,146 +1,129 @@
 import * as React from 'react';
-import makeStyles from '@mui/styles/makeStyles';
-import Select from '@mui/material/Select';
-import MenuItem from '@mui/material/MenuItem';
-import { SelectChangeEvent } from '@mui/material';
-import TextField from '@mui/material/TextField';
-import Tooltip from '@mui/material/Tooltip';
-import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
 import { useHistory } from 'react-router-dom';
+import makeStyles from '@mui/styles/makeStyles';
+import MenuItem from '@mui/material/MenuItem';
+import Paper from '@mui/material/Paper';
+import Box from '@mui/material/Box';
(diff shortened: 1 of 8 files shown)


client/src/components/Search.tsx
```diff
@@ -1,146 +1,99 @@
...
+    (querySearchCategory as string) || 'Needs',
+  );
+  const [searchText, setSearchText] = React.useState<string>((querySearchText as string) || '');
 
-  const handleKeyDown = (e: React.KeyboardEvent) => {
-    if (e.key === 'Enter') {
-      history.push(`/SearchResults?search=${searchText}&category=${selectedSearchCategory}`);
-    }
+  const handleSearch = () => {
+    history.push(`/SearchResults?search=${searchText}&category=${searchCategory}`);
```
esteban-gs, line 43: Not part of your work, but if we can change the route here and in the router to `search-results` instead the link would be more semantic with typical kabab-cased routes
```suggestion
    history.push(`/search-results?search=${searchText}&category=${searchCategory}`);
```

client/src/components/Search.tsx
```diff
@@ -1,146 +1,99 @@
...
 
-  const handleDropdownChange = (event: SelectChangeEvent<string>) => {
-    setSelectedSearchCategory(event.target.value as string);
-  };
+  const searchParams = new URLSearchParams(history.location.search);
+  const querySearchText = searchParams.get('search');
+  const querySearchCategory = searchParams.get('category');
+
+  const [searchCategory, setSearchCategory] = React.useState<string>(
+    (querySearchCategory as string) || 'Needs',
```
esteban-gs, line 38: It will be safer to `null` coalesce - string casting is not checked during run-time
```suggestion
  const [searchCategory, setSearchCategory] = React.useState<string>(
   querySearchCategory ?? 'Needs',
```

client/src/components/Search.tsx
```diff
@@ -1,146 +1,99 @@
...
-          variant="outlined"
+          onKeyDown={(e) => e.key === 'Enter' && handleSearch()}
+          onChange={(e: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) => {
+            setSearchText(e.target.value);
+          }}
         />
-      </Tooltip>
+        <IconButton onClick={handleSearch} sx={{ paddingRight: '10px', marginLeft: 'auto' }}>
+          <SearchIcon fontSize="large" />
+        </IconButton>
```
esteban-gs, line 95: Maybe let's disable is the search string is empty

esteban-gs (approved):
Thanks for cleaning this up. Just a couple of comments, but otherwise, it looks good

PR #301, opened 2023-03-16
Add categories to org

rmwerner18:


jd2rogers2:
lgtm so far but you can give me some test cases?
maybe using the new swagger console (localhost:3001/docs)
something like
if i hit the create org endpoint with categories: [blah] data i should succeed (see category in GET response)
if i hit the create org endpoint with categories: [not_blah] data i should fail the validator
cc: @esteban-gs

rmwerner18:
@jd2rogers2 I haven't written tests before but gave it a shot, let me know how I did/what I need to change!

esteban-gs (approved):
LGTM! Should be good to go after you solve the merge conflicts

Issue #64, opened 2021-10-27
make migrations manual


jd2rogers2:
ACs:
- https://github.com/Nonprofit-Exchange-Hub/web-app/blob/c8a1d6b2c6e3894b243e4f926444e81155eb38e5/server/src/database-connection.service.ts#L15 should be set to false so that we don't automatically clear previous tables every time we start a new connection
- use resources below to set up a new workflow for when we update entity schemas
  - keep in mind that this new workflow will also be utilized in deployment build scripts
- It should be enough to have the `typeorm`-generated sql transactions there
- (DONE) provide some seed data (arguably separate PR, let's chat if this ends up being difficult)
  - https://typeorm.io/#/insert-query-builder
- update set up instructions in readme to reflect new migration step
  - and maybe some how to documentation for day to day use of these as well

### Acceptance criteria
- we should have manual typeorm migrations in place that we can run from a script. similar to the `seed:drop` script
- Optional: add the script to a process on the docker-compose for development, so that the migration runs on `docker-compose up`
- We can see a clear flow of creating a new database, running the migrations and seeding the db

resources:
https://docs.nestjs.com/techniques/database#migrations
https://typeorm.io/#/migrations
https://typeorm.io/#/using-cli

Blocks: https://github.com/Nonprofit-Exchange-Hub/web-app/issues/308

PR #303, opened 2023-03-18
Searchable assets

skyfox93:
## Dev Summary

Enables full text search of assets.
This allows:
- matches when query is missing a word e.g. 'three doves' now returns the result 'three turtle doves' when previously it would not
- ranking of matches in order of relevance (they were previous not ranked)

(This was done out of interest even though not a priority, I had mentioned looking at full text search in an interview and wanted to follow through)

* Further Notes/ Explaination
Best practice appears to be to use a generated ts_vector column with a GIN Index in as full text search can be slow otherwise.

I did some benching marking using another database with 20,000 entries and casting to a ts_vector ( without a sperate column) took around 750 ms, searching with a ts_vector column took around 40 ms, and searching with an gin index on a tsvector field around 1ms.
 
* Further Steps

Eventually, since our app has an option to query multiples entities at once, we may want to move searching all of these entities into a single query. 
And example of a potential strategy can be found here : https://thoughtbot.com/blog/implementing-multi-table-full-text-search-with-postgres 

## Test Plan

repro steps:

1. localhost:3000 
2. search needs for "three doves" 
3. "three turtle doves" is among the results ( it previously was not)

## Resources


esteban-gs:
@skyfox93 , thanks for putting this together. This looks like it's going to be a good path forward. Before merging this, we should implement migrations, referenced in https://github.com/Nonprofit-Exchange-Hub/web-app/issues/64

server/sql-scripts/migrations/1679101636074-addsearchtitleindex.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+            await queryRunner.query(`CREATE INDEX "searchtitleindex" ON "assets" USING GIN("searchtitle")`)
+            
+        }
+    
+        async down(queryRunner: QueryRunner): Promise<void> {
+            await queryRunner.query(
+                `drop index "searchtitleindex"`,
+            ) // reverts things made in "up" method
```
esteban-gs, line 13: ok to remove this comment

server/sql-scripts/migrations/1679101636074-addsearchtitleindex.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class addsearchtitleindex1679101636074 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+            await queryRunner.query(`CREATE INDEX "searchtitleindex" ON "assets" USING GIN("searchtitle")`)
+            
+        }
```
esteban-gs, line 8:
```suggestion
    public async up(queryRunner: QueryRunner): Promise<void> {
            await queryRunner.query(`CREATE INDEX "searchtitleindex" ON "assets" USING GIN("searchtitle")`)
        }
```

server/src/assets/assets.service.ts
```diff
@@ -22,19 +22,30 @@ export class AssetsService {
...
-      }) || []
-    );
+    const { limit, offset, search, ...rest } = getAssetsDto;
+    if (search) {
+      return this.assetsRepository
+        .createQueryBuilder()
+        .select('asset')
+        .from(Asset, 'asset')
+        .where(`"asset"."searchtitle" @@ plainto_tsquery('english', :query)`, {
+          query: `${search}`,
```
esteban-gs, line 32: Do you know if Typeorm sanitizes this `query` property? if not we should add at least a basic sanitation function with `toLower` `trim` and replace any instances of `drop` with blank text.

server/src/assets/dto/update-asset.dto.ts
```diff
@@ -1,5 +1,9 @@
 import { PartialType } from '@nestjs/mapped-types';
+import { Exclude } from 'class-transformer';
 
 import { CreateAssetDto } from './create-asset.dto';
 
-export class UpdateAssetDto extends PartialType(CreateAssetDto) {}
+export class UpdateAssetDto extends PartialType(CreateAssetDto) {
+  @Exclude()
+  searchtitle: string;
```
esteban-gs, line 8: I don't believe we need to exclude this here, since the return type is the `entity.asset.ts`

server/sql-scripts/migrations/1679101636074-addsearchtitleindex.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class addsearchtitleindex1679101636074 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+            await queryRunner.query(`CREATE INDEX "searchtitleindex" ON "assets" USING GIN("searchtitle")`)
```
esteban-gs, line 6: should we apply tim and tolower on `searchtitle`?

esteban-gs (changes requested):
Looks good. I left a couple of thoughts

esteban-gs (approved):

PR #309, opened 2023-03-26
fix profile route

esteban-gs:
## Dev Summary

Just cleaning up the profile route

## Test Plan

repro steps:

1. localhost:3000 
2. start ui and login with a valid user credential
expected: 

- when you click on the `My Profile` link, the route reads `/profile` instead of `/users/:userId`

## Resources

n/a

### DIFF
--- a/client/src/routes/routes.ts
+++ b/client/src/routes/routes.ts
@@ -115,7 +115,7 @@ const routes: RouteMap = {
   User: {
     component: User,
     roles: ['OWNER', 'ADMIN'],
-    path: '/profile',
+    path: '/profile/:id?',
   },
   ActionForm: {
     component: ActionForm,
--- a/client/src/routes/routes.ts
+++ b/client/src/routes/routes.ts
@@ -115,7 +115,7 @@ const routes: RouteMap = {
   User: {
     component: User,
     roles: ['OWNER', 'ADMIN'],
-    path: '/users/:userId',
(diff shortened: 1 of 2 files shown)


client/src/routes/routes.ts
```diff
@@ -115,7 +115,7 @@ const routes: RouteMap = {
   User: {
     component: User,
     roles: ['OWNER', 'ADMIN'],
-    path: '/users/:userId',
+    path: '/profile',
```
jd2rogers2, line 118: i think we'll want to show other users', not just the current user, profile page. so i think this isn't the right fix
we should instead fix the link to point to /users/[current_user.id]

jd2rogers2 (changes requested):

client/src/routes/routes.ts
esteban-gs: So, the `user-id` path param determines the public user profile to fetch?

jd2rogers2: that's how i imagined it/the most common pattern i've seen. if you have another you prefer then that's fine with me

jd2rogers2: Sure, so i guess what you had originally is fine

So both urls point to the same react component
logic in there would be: 
Use url param user ID
If doesn't exist, use current user

jd2rogers2 (approved):

PR #318, opened 2023-04-09
add email_verified: true

taoliu12:
added "email_verified: true," to both seed users, and Login works correctly now.


esteban-gs (changes requested):
Nice! this should get us closer to closing this issue. A couple of changes in comments and here:

- At one point we added the `email_verified` property back to the `create-user-dto.ts` class. This should be reverted. Even though the card doesn't require if explicitly, we could benefit from fixing this. 
- Requested changes:
  - Remove the `email_verified?` property from `create-user.dto` - `src/acccount-manager/dto/create-user.dto.ts`
  - In the seed date. use the `CreateUserInternal` - `src/acccount-manager/dto/create-user.internal.ts` type instead of `CreateUserDto`. 

The main idea behind this is to disallow the controller's ability to update the `email_verified` field, since it controls access to the app. Internally, we can still override it by using the service type `CreateUserInternal` which has access to that property.

server/src/acccount-manager/dto/create-user.dto.ts
taoliu12, line 1: Sorry, not sure why the entire file is diff'ed.
But I removed the 
`  @IsBoolean()
  email_verified?: boolean;`

taoliu12:
@esteban-gs I made the requested changes, and re-seeded to make sure login still works.  

Also, sorry, not sure why the entire file is diff'ed .
But I removed the email_verified? property from create-user.dto

esteban-gs (approved):
Looks good to me!

PR #320, opened 2023-04-18
User profile page

rmwerner18:
Currently, this assumes that users can only see their own profiles--I figured we would change this when we change the route from '/profile/:id' to having an actual id. Also, none of the buttons work, at this point it's all just styling.


client/src/views/User.tsx
```diff
@@ -1,13 +1,133 @@
-import { useParams } from 'react-router-dom';
+import React from 'react';
+// import { useParams } from 'react-router-dom';
```
esteban-gs, line 2: We can probably remove this comment?

esteban-gs (approved):
Looks good to me, except for a small comment

PR #324, opened 2023-04-22
run npm install only when there are changes to package.json

skyfox93:
runs npm install only when there are changes to package.json


esteban-gs (approved):
Nice! This will speed up our dev. Just on comment, but should be ready to merge once you fix it

PR #327, opened 2023-04-26
290 user sign up form validation

yulejessica:
## Dev Summary

The user sign-up form should validate accepted terms, email, and zip and provide useful validation errors to the user for each of those fields.
This will require using a form library (I used the UserOrg sign-up form as a reference ) as well as a multi-step verification form (when the user can go to the next step)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


client/src/components/Users/Auth/SignUpCitizen/SignUpCitizen.tsx
```diff
@@ -0,0 +1,590 @@
...
+      setUser(data);
+      handleNext();
+    }
+  };
+
+  return (
+    <div className="SignupCitizen">
+      {/* <p>dirtyFields: {JSON.stringify(dirtyFields)}</p>
+      <p>isValid: {isValid.toString()}</p> */}
+      <Grid container>
```
esteban-gs, line 141:
```suggestion
  return (
    <div className="SignupCitizen">
      <Grid container>
```

esteban-gs, line 141:
```suggestion
  return (
    <div className="SignupCitizen">
      <Grid container>
```

client/src/components/Users/Auth/SignUpCitizen/validation-rules.ts
```diff
@@ -0,0 +1,43 @@
...
+  vals: FormData,
+): boolean => {
+  const errVals = Object.values(errors);
+  const valsAsStringArr = Object.values(vals);
+  return (
+    errVals.every((x) => x === undefined) &&
+    valsAsStringArr.length === 17 &&
+    valsAsStringArr.every((s) => !!s)
+  );
+};
```
esteban-gs, line 43: We no longer need this one, right? It's not used anywhere else
```suggestion
```

client/src/views/SearchResults.tsx
```diff
@@ -128,7 +128,7 @@ function SearchResults(): JSX.Element {
 
   React.useEffect(() => {
     fetchSearchData();
-  }, [querySearchText, querySearchCategory]);
+  }, [querySearchText, querySearchCategory, fetchSearchData]);
```
esteban-gs, line 131: I don't think we need `fetchSearchData` here right?

client/src/views/TermsOfService.tsx
esteban-gs, line 1: @yulejessica , are the style changes to this component necessary?

esteban-gs (changes requested):

client/src/views/TermsOfService.tsx
yulejessica: Probably not, but I added to make the footer not stick to the top.

client/src/views/SearchResults.tsx
esteban-gs: ok. I think it's ok to leave it as you have it then

esteban-gs (approved):
Looks good!

PR #334, opened 2023-04-30
include email_opt_out in new user from userOrg signup

rmwerner18:
email_notification_opt_out wasn't being sent to the back end with org sign up data. Fixed this, and everything else looks good.


esteban-gs (approved):
Nice!

PR #336, opened 2023-05-04
Add to swagger API documentation

marvinsjsu:
## Dev Summary

This is one of possibly 3 PRs to add to the Swagger API documentation.  This PR has updates to `account-manager`, `assets`, and `categories` controllers and dtos to expose information regarding API requests and responses.

## Test Plan

repro steps:

1. localhost:3001/docs 
2. look through and verify the details displayed in `assets`, `auth`, `categories` endpoints
3.  test login, password reset, and CRUD operations for `assets` and `categories`

When testing requests on Swagger, we'll need to authorize first by copying and pasting `NEH_is_cool` cookie value:
- run `localhost:3000` and login via `http://localhost:3000/login`
- open dev console and go to `Application` tab
- in `Storage` section, click on `Cookies` (should now see `NEH_is_cool`)
Screenshot 2023-05-04 at 11 38 50 AM
- in `localhost:3001/docs`, click on `Authorize` button and enter the cookie value
Screenshot 2023-05-04 at 11 40 02 AM

## Resources

- [NextJS Swagger documentation](https://docs.nestjs.com/openapi/introduction)


esteban-gs (changes requested):
@marvinsjsu , thanks or working on this! Looking good so far. I don't see anything incorrect with your changes, but with how you approached this, I think it would be a good opportunity to decide on one way of getting OpenApi to generate our Api docs.

There are 2 approaches to to decide from.
- Rely on nestjs auto-annotate plugin (my personal preference), it requires a heavy use of `.dto.ts` classes and keep the `@Api...` annotations to a minimum. A downside of this is that we have to duplicate the return type properties, but this might scale better if when we make a heavier use of `class-transformer` if we ever do any formatting before sending to the client. I think the card left out that we set up OpenApi to auto-annotate on app build and it will pick up any files with the .dto suffix, so really, we don't need these annotations in the dtos.

You can see where we configure this auto-annotate feature in nest-cli.json.

- Annotate each entity field and each controller method. This is a more manual approach, and requires a lot of annotations, but it does remove the need to create a `.dto.ts` return type

server/src/acccount-manager/account-manager.controller.ts
```diff
@@ -207,7 +241,10 @@ export class AccountManagerController {
       limits: { fileSize: FileSizes.MB },
     }),
   )
-  @ApiResponse({ status: 200, type: ReturnUserDto })
+  @ApiOkResponse({
+    description: 'Uploaded profile image.',
+    type: ReturnUserDto,
+  })
```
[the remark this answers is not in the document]
marvinsjsu: Thanks for pointing these out.  I also prefer letting the auto annotation plugin handle the work as it was pretty tedious to do this manually.

server/src/acccount-manager/dto/create-user.dto.ts
```diff
@@ -1,36 +1,48 @@
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+
 import { IsEmail, IsNotEmpty, IsBoolean, IsOptional, Validate } from 'class-validator';
 import { InterestsProps, InterestNamesIsArray } from '../../shared/interests.validator';
 import { Interests } from '../../shared/interests.dto';
 
 export class CreateUserDto {
+  @ApiProperty()
```
esteban-gs, line 8: In this dto file, we can remove all the `@ApiProperty()` annotations from `@nestjs/swagger`. The OpenApi plugin will pick up the required/optional properties as well as the interest type based on the typescript types.

esteban-gs (changes requested):
@marvinsjsu Thank you for  working on this. Looks good. I have one comment so far where we can concretely remove annotations

server/src/acccount-manager/entities/user.entity.ts
esteban-gs, line 1: @marvinsjsu , we can also remove all instances of `User` as the return type in `account-manager.controler.ts` and user the existing `ReturnUserDto` class instead as a return type. 

Once you do that, you can remove all the OpenApi annotations from this class.

server/src/assets/dto/create-asset.dto.ts
esteban-gs, line 1: You can also remove all the instances of `@ApiProperty()` and `@ApiPropertyOptional()` from this dto since the OpenApi can infer this from the types and the `foo?: string` optional property operator.

server/src/assets/dto/get-asset.dto.ts
esteban-gs, line 1: You can safely remove the annotations here as well.

server/src/assets/entities/asset.entity.ts
esteban-gs, line 1: Instead of annotating this entity, we should create a new class following the dto convention, similar to `ReturnUserDto`. Inheriting from this class won't work for auto-documenting, so you'll have to declare the same properties as this class. We can trade off auto-documentation for code repetition.

server/src/categories/dto/create-category.dto.ts
esteban-gs, line 1: @marvinsjsu , the annotations here can be removed as well

server/src/categories/dto/get-categories-filter.dto.ts
esteban-gs, line 1: Instead of extending from the `CreateCategoryDto` let's declare the properties here even if they are duplicated, This will also be useful when we use `class-transformer`.

This way, dtos can be the place where we deal with class-transformations and return types and it's decoupled from entities.

server/src/categories/entities/category.entity.ts
```diff
@@ -1,17 +1,22 @@
 import { Entity, Column, PrimaryGeneratedColumn, Index } from 'typeorm';
+import { ApiProperty } from '@nestjs/swagger';
 
 @Entity('categories')
 export class Category {
+  @ApiProperty()
   @PrimaryGeneratedColumn()
   id: number;
```
esteban-gs, line 8: You can get the same behavior by removing the annotation and adding the optional property denotator:

```suggestion
  @PrimaryGeneratedColumn()
  id: number?;
```

server/src/categories/dto/create-category.dto.ts
```diff
@@ -1,12 +1,16 @@
...
   @IsNotEmpty()
   name: string;
 
+  @ApiPropertyOptional()
   @IsOptional()
   applies_to_assets: boolean;
 
+  @ApiPropertyOptional()
   @IsOptional()
   applies_to_organizations: boolean;
```
esteban-gs, line 15: You can get the same behavior by removing the annotation and adding the optional property denotator:

```suggestion
  @IsOptional()
  applies_to_assets?: boolean;

  @IsOptional()
  applies_to_organizations?: boolean;
```

marvinsjsu:
@esteban-gs, sorry it took me some time to get back to working on this. Hope the changes align to a better approach, and thank you for your patience on this.

server/src/assets/dto/return-asset.dto.ts
esteban-gs, line 1: 👍🏾

server/src/acccount-manager/entities/user.entity.ts
```diff
@@ -7,36 +8,47 @@ import { Transaction } from '../../transactions/entities/transaction.entity';
 
 @Entity('users')
 export class User {
+  @ApiProperty()
```
esteban-gs, line 11: @marvinsjsu , these can be removed as well

server/src/acccount-manager/dto/create-user.dto.ts
```diff
@@ -1,36 +1,48 @@
...
 import { IsEmail, IsNotEmpty, IsBoolean, IsOptional, Validate } from 'class-validator';
 import { InterestsProps, InterestNamesIsArray } from '../../shared/interests.validator';
 import { Interests } from '../../shared/interests.dto';
 
 export class CreateUserDto {
+  @ApiProperty()
   @IsNotEmpty()
   firstName: string;
 
+  @ApiProperty()
```
esteban-gs, line 12: These can be removed

esteban-gs (changes requested):

esteban-gs:
@marvinsjsu  looking good! Thanks for working through those changes.

esteban-gs (approved):
@marvinsjsu , thanks for working though these changes

marvinsjsu:
@esteban-gs, I very much appreciate your patience on this and the follow-up PR.

PR #337, opened 2023-05-04
Sign out header new layout 

paaaarv:
## Dev Summary
 Figma link for signed out header: https://www.figma.com/file/K7DT5jYcLKRc9epmNrO0N7/UX---Dev-Handoff?node-id=200-5761&t=0a5133SJk6atcWfw-0

Changed header to match Figma link when user is signed out


client/src/components/Header.tsx
```diff
@@ -52,6 +57,25 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    width: '100px',
+  },
+  signIn: {
+    backgroundColor: 'white!important',
+    color: '#323232!important',
+    borderRadius: '10px',
+    border: '1px solid #323232!important',
+    marginLeft: '5px!important',
+    width: '100px',
+  },
```
esteban-gs, line 78: @paaaarv , is there a way to use MUI props to achieve this style, similar to this: https://github.com/Nonprofit-Exchange-Hub/web-app/commit/4f732a051319f326ae696f47afae383e3a18815f#diff-b36271bb100b70cfb099cab9923f9dba10e0476f6f67d821a09cc879afa185ff 

If not, then this should be fine.

esteban-gs (approved):
@paaaarv , thanks for working on this! I just have one comment on using a different approach. Otherwise, it looks good to me!

client/src/components/Header.tsx
paaaarv: hey! do you mean passing in any styles to sx props?

PR #341, opened 2023-05-13
Add to swagger API documentation (part 2)

marvinsjsu:
## Dev Summary

Follow-up to [PR#336](https://github.com/Nonprofit-Exchange-Hub/web-app/pull/336).  This PR has updates to `message`, `user-org`, and `transaction` controllers and dtos to expose information regarding API requests and responses.

## Test Plan

repro steps:

1. localhost:3001/docs 
2. look through and verify the details displayed in `message`, `user-org`, and `transaction` endpoints
3.  test login, password reset, and CRUD operations for `assets` and `categories`

When testing requests on Swagger, we'll need to authorize first by copying and pasting `NEH_is_cool` cookie value:
- run `localhost:3000` and login via `http://localhost:3000/login`
- open dev console and go to `Application` tab
- in `Storage` section, click on `Cookies` (should now see `NEH_is_cool`)
Screenshot 2023-05-04 at 11 38 50 AM
- in `localhost:3001/docs`, click on `Authorize` button and enter the cookie value
Screenshot 2023-05-04 at 11 40 02 AM

## Resources

- [NextJS Swagger documentation](https://docs.nestjs.com/openapi/introduction)


marvinsjsu:
@esteban-gs, I've updated this PR to add return dtos instead of using the annotations on the entities as well.  Let me know what you think.

Also, I'm curious ... is there any diagram that shows the flow of how a user creates a transaction?  Would a user need to be associated with an organization before they can create an asset?

server/src/organizations/dto/return-organization.dto.ts
```diff
@@ -0,0 +1,8 @@
+import { IsNotEmpty } from 'class-validator';
+
+import { CreateOrganizationDto } from './create-organization.dto';
+
+export class ReturnOrganizationDto extends CreateOrganizationDto {
+  @IsNotEmpty()
+  id: number;
+}
```
esteban-gs, line 8: I think we should define the properties here instead of extending from `createDto..`. It should be most properties defined in the entity, except any sensitive ones.

server/src/messages/messages.service.ts
```diff
@@ -6,12 +6,13 @@ import { User } from '../acccount-manager/entities/user.entity';
...
 import { CreateMessageDto } from './dto/create-message.dto';
 import { UpdateMessageDto } from './dto/update-message.dto';
+import { ReturnUserDto } from '../acccount-manager/dto/auth.dto';
 
 @Injectable()
 export class MessagesService {
   constructor(@InjectRepository(Message) private messagesRepository: Repository<Message>) {}
 
-  async create(createMessageDto: CreateMessageDto, user: User): Promise<Message> {
+  async create(createMessageDto: CreateMessageDto, user: ReturnUserDto): Promise<Message> {
```
esteban-gs, line 15: I think we should be able to keep the User type here, since it's a service.

server/src/organizations/organizations.controller.ts
esteban-gs, line 1: 💯

esteban-gs (approved):
@marvinsjsu , I just have a couple more comments, but overall, it looks good to me

server/src/messages/messages.service.ts
marvinsjsu: @esteban-gs, I've reverted my change so it uses `User` type, but I am curious about the mismatch in the types  of params we're using for this service, `CreateMessageDto` and `User`.  My initial thought was that using `dtos` would be more inline with how data is moved around the application, but I'm new to Nest.js and so might not be understanding this correctly.

server/src/organizations/dto/return-organization.dto.ts
marvinsjsu: @esteban-gs, my thought was that data in the `createDto..`, the properties correspond to data provided by the user and so can be exposed, but I see that there might be user values we want to hide.  I was thinking the following properties could be sensitive: `nonprofit_classification` and `ein`.  Are there other properties we should be excluding in the response?

PR #342, opened 2023-05-18
Match back and front end user signup

rmwerner18:
-Made bio, zip code, and interests inputs functional and included data in post request
-Fixed a bug that was happening with the city/state select inputs 
-Post request to create user works (successfully creates user) but still throws an error for some reason
-Profile_image_url is not included in post request, currently image input is a file upload and not a text field for a url. This will probably require some more steps in order to store file as url
-Added interests from frontend to category seeds in backend, so that these interests can be saved 
-Commented out email_verified check when logging in, to allow us to work around that for now

### DIFF
--- a/client/src/components/Users/Auth/SignUpCitizen/SignUpCitizen.tsx
+++ b/client/src/components/Users/Auth/SignUpCitizen/SignUpCitizen.tsx
@@ -74,8 +74,12 @@ function SignupCitizen() {
 
   const makeCitySelectOptions = (state: string) => {
     // TODO: get list of cities from state
-    let cities = Array.from(Array(50).keys()).map((num) => <MenuItem value={num}>{num}</MenuItem>);
-    return cities;
+    const citiesArray = ['Seattle', 'Detroit', 'Chicago', 'Nashville'];
+    return citiesArray.map((city) => {
+      return <MenuItem value={city}>{city}</MenuItem>;
+    });
+    // let cities = Array.from(Array(50).keys()).map((num) => <MenuItem value={num}>{num}</MenuItem>);
+    // return cities;
   };
 
   const handleChange = (evt: React.ChangeEvent<HTMLInputElement>): void => {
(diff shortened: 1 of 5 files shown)


client/src/components/Users/Auth/SignUpCitizen/SignUpCitizen.tsx
```diff
@@ -74,8 +90,10 @@ function SignupCitizen() {
 
   const makeCitySelectOptions = (state: string) => {
     // TODO: get list of cities from state
-    let cities = Array.from(Array(50).keys()).map((num) => <MenuItem value={num}>{num}</MenuItem>);
-    return cities;
+    const citiesArray = ['Seattle', 'Detroit', 'Chicago', 'Nashville'];
+    return citiesArray.map((city) => {
+      return <MenuItem value={city}>{city}</MenuItem>;
+    });
```
esteban-gs, line 96: @rmwerner18 , for now, let's make this field a simple input element similar to zipcode. This will simplify our work.

client/src/components/Users/Auth/SignUpCitizen/SignUpCitizen.tsx
```diff
@@ -297,14 +317,16 @@ function SignupCitizen() {
...
+                        value={formData.state}
                       >
-                        {makeCitySelectOptions('WA')}
+                        {makeStateSelectOptions()}
                       </Select>
                     </Grid>
                     <Grid item xs={6}>
+                      <InputLabel id="city-select">City</InputLabel>
                       <Select
+                        labelId="city-select"
```
esteban-gs, line 329: Let's make this a regular input instead, so we don't have to provide a city for each state.

esteban-gs (approved):
@rmwerner18 , thanks for working on this! I made a comment on changing one field, but otherwise it looks good to me. Approved once you apply the recommended change and remove the comment on the controller.

PR #346, opened 2023-05-25
minor fixes - add keys to maps

marvinsjsu:
## Dev Summary

This PR has fixes for warnings regarding the need of keys when using `.map()`.  This should help reduce the noise in our browser console.

Screenshot 2023-05-25 at 12 06 40 AM

## Test Plan

There aren't any functional changes with the PR.  That said, we can still go through the app for manual testing.


esteban-gs (approved):
Thanks for fixing this, it LGTM! 🥏

PR #347, opened 2023-05-30
135 server env configs

esteban-gs:
## Dev Summary

This PR adds better management of our app env vars. I added an extra env var to toggle logging on and off and added extensibility for `development` and `staging` environments. 

I added a app configs validation step that provides some useful info.

## Test Plan

This will require changing some configs in our staging environments and changing devs' local `.env` files.

repro steps:

1. start the express server
2. expect the app to work
3. If you set an invalid value for `NODE_ENV`, expect to see some helpful error messages

## Post Merge Tasks
- [ ] change staging environment variables to match changes here
- [ ] post in slack to let devs know about the changes

### DIFF
--- a/server/sample.env
+++ b/server/sample.env
@@ -1,6 +1,6 @@
 
 # if you change DATABASE_DB, also change db name in init.sql
-MODE=dev
+NODE_ENV=development
 PORT=3001
 DATABASE_HOST=localhost
 DATABASE_PORT=5432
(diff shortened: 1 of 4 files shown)


server/sample.env
```diff
@@ -1,6 +1,6 @@
 
 # if you change DATABASE_DB, also change db name in init.sql
-MODE=dev
+NODE_ENV=development
```
esteban-gs, line 3: Changing this to match the standard express env var

server/sample.env
```diff
@@ -11,6 +11,7 @@ POSTGRESQL_SSL_CA=""
 POSTGRESQL_SSL_CERT=""
 POSTGRESQL_SSL_KEY=""
 BCRYPT_WORK_FACTOR=10
+DB_LOGGING=true
```
esteban-gs, line 14: We can toggle this on/off in our different environments to debug DB queries

taoliu12 (approved):
ran it on my machine, and it looks good to me.

PR #348, opened 2023-06-08
Feature 288 org sign up screens

denniswangcodes:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


denniswangcodes:
Would appreciate insight on:
-turn “Already have an account?”’s Sign Up button to clickable text

Currently fine-tuning: 
-able to uncheck accept terms & services checkbox
-restrict next/submit if field left blank
-add check EIN function injection

esteban-gs:
@denniswangcodes , I believe the build error will go away once you update your branch with the main branch.

marvinsjsu:
hi @denniswangcodes, it looks like we might have a small typo in `SignupNonProfit.tsx`

```
ERROR in src/views/SignupNonProfit.tsx:22:28
TS1261: Already included file name '/Users/marvin/Projects/external/web-app/client/src/views/FocusAreas.tsx' differs from file name '/Users/marvin/Projects/external/web-app/client/src/views/focusAreas.tsx' only in casing.
  The file is in the program because:
    Imported via './FocusAreas' from file '/Users/marvin/Projects/external/web-app/client/src/views/SignupNonProfit.tsx'
    Root file specified for compilation
    20 | import { ModalContext } from '../providers/ModalProvider';
    21 | import { useForm } from 'react-hook-form';
  > 22 | import { focusAreas } from './FocusAreas';
```

marvinsjsu:
TypeScript is also complaining about the expected prop types for the `Select` element.

```
ERROR in src/components/Users/Auth/SignUpUserAndNonprofit/SignUpUserAndNonprofit.tsx:495:28
TS2322: Type '{ children: (Element | Element[])[]; placeholder: string; variant: "outlined"; autoWidth: true; input: Element; inputProps: { 'aria-label': string; }; displayEmpty: true; ... 4 more ...; ref: RefCallBack; }' is not assignable to type 'SelectProps<string>'.
  Types of property 'onChange' are incompatible.
    Type '(event: string | ChangeEvent<Element>) => void' is not assignable to type '(event: SelectChangeEvent<string>, child: ReactNode) => void'.
      Types of parameters 'event' and 'event' are incompatible.
        Type 'SelectChangeEvent<string>' is not assignable to type 'string | ChangeEvent<Element>'.
          Type 'Event & { target: { value: string; name: string; }; }' is not assignable to type 'string | ChangeEvent<Element>'.
            Type 'Event & { target: { value: string; name: string; }; }' is missing the following properties from type 'ChangeEvent<Element>': nativeEvent, isDefaultPrevented, isPropagationStopped, persist
    493 |                         <>
    494 |                           <FormLabel>IRS Nonprofit Organization Classification</FormLabel>
  > 495 |                           <Select
        |                            ^^^^^^
    496 |                             {...field}
    497 |                             placeholder="Select classification"
    498 |                             variant="outlined"
```

marvinsjsu:
Do you know if we have a list of required fields for the signup?  Asking because we may need to create a follow-up ticket to add that validation.  Also, about the last step.  It shows the `Sign Up` button after the user has gone through the workflow - do we need to ask design about removing this? 

Screenshot 2023-09-06 at 9 49 54 AM

client/src/components/Modals/SignUpModal.tsx
```diff
@@ -1,5 +1,16 @@
-// import Button from '@mui/material/Button';
 import React from 'react';
+import routes from '../../routes/routes';
+import Grid from '@mui/material/Grid';
```
marvinsjsu, line 3: I'm wondering if it's possible to consolidate the MUI imports, like so ...
```
import {
  Grid,
  Dialog,
  Button,
  ...
} from '@mui/material';
```

Also from what I've seen in other projects (and we don't have to follow it), but it might be nice to arrange imports based on if the import is for a library or package, then followed by internal components, then images, then styling.  We just separate the set by a space, so for example ...

```
import React from 'react';
import { Link } 'react-router-dom';
import {
  Grid,
  Dialog,
  ...
} from '@mui/material';

import routes from '../../routes/routes';
```

 Let me know what you think.  Thanks.

client/src/views/Signup.tsx
```diff
@@ -37,39 +46,49 @@ const useStyles = makeStyles()((theme: Theme) => ({
...
-      </Typography>
-      <Typography variant="h2" component="h2">
-        Select your account type.
-      </Typography>
-      <Typography variant="body2" component="caption">
-        Already have an account? <Link to={routes.Login.path}>Log In</Link>{' '}
-      </Typography>
+  const [open, setOpen] = React.useState(true);
+  const onClose = (e: any, reason: string) => {
+    if (reason !== 'backdropClick') {
```
marvinsjsu, line 51: I'm curious about `backdropClick` - I'll need to read up on this :)

client/src/views/SignupNonProfit.tsx
```diff
@@ -1,7 +1,836 @@
...
+import * as React from 'react';
+import Grid from '@mui/material/Grid';
+import Input from '@mui/material/Input';
+import Button from '@mui/material/Button';
+import Checkbox from '@mui/material/Checkbox';
+import Typography from '@mui/material/Typography';
+import FormControl from '@mui/material/FormControl';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import type { Theme } from '@mui/material/styles';
+import { useContext } from 'react';
```
marvinsjsu, line 18: same idea re: consolidating the imports ... we can probably replace the top line with 
```
import React, { useContext, useEffect } from 'react';
```

client/src/views/SignupNonProfit.tsx
```diff
@@ -1,7 +1,836 @@
...
+  const [disableNext, setDisableNext] = React.useState<boolean>(true);
+  const [activeStep, setActiveStep] = React.useState<number>(0);
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+  const [emailError, setEmailError] = React.useState<string>('');
+  const [formData, setFormData] = React.useState(initialFormData);
+  const { user, setUser } = React.useContext(UserContext);
+  const modalContext = useContext(ModalContext);
+  const { openModal } = modalContext;
+  const { control } = useForm();
+  const steps = [
```
marvinsjsu, line 153: This can probably be placed outside the function, since this doesn't change.

marvinsjsu: This component is pretty big.  It might be good idea to refactor this into multiple smaller components and this can be a parent or container component.

marvinsjsu (commented):
This is looking good.  There are a couple of issues regarding a filename and TS, and I've also added a couple of suggestions.  Otherwise, nice work!

esteban-gs:
> TypeScript is also complaining about the expected prop types for the `Select` element - I was able to bypass this by adding `{/* @ts-ignore */}` right above line 495, `<Select>`.
> 
> ```
> ERROR in src/components/Users/Auth/SignUpUserAndNonprofit/SignUpUserAndNonprofit.tsx:495:28
> TS2322: Type '{ children: (Element | Element[])[]; placeholder: string; variant: "outlined"; autoWidth: true; input: Element; inputProps: { 'aria-label': string; }; displayEmpty: true; ... 4 more ...; ref: RefCallBack; }' is not assignable to type 'SelectProps<string>'.
>   Types of property 'onChange' are incompatible.
>     Type '(event: string | ChangeEvent<Element>) => void' is not assignable to type '(event: SelectChangeEvent<string>, child: ReactNode) => void'.
>       Types of parameters 'event' and 'event' are incompatible.
>         Type 'SelectChangeEvent<string>' is not assignable to type 'string | ChangeEvent<Element>'.
>           Type 'Event & { target: { value: string; name: string; }; }' is not assignable to type 'string | ChangeEvent<Element>'.
>             Type 'Event & { target: { value: string; name: string; }; }' is missing the following properties from type 'ChangeEvent<Element>': nativeEvent, isDefaultPrevented, isPropagationStopped, persist
>     493 |                         <>
>     494 |                           <FormLabel>IRS Nonprofit Organization Classification</FormLabel>
>   > 495 |                           <Select
>         |                            ^^^^^^
>     496 |                             {...field}
>     497 |                             placeholder="Select classification"
>     498 |                             variant="outlined"
> ```

@denniswangcodes , @marvinsjsu  we came across this in another PR. This might help:

https://github.com/Nonprofit-Exchange-Hub/web-app/pull/355/files#diff-58627174d7ef27a96ddf1421e598335d42284e257344118545e48afd88c4c753R497-R503

client/src/views/focusAreas.tsx
esteban-gs, line 1: @denniswangcodes , this file's extension should be `.ts` instead of `.tsx`.

package.json
esteban-gs, line 1: This along with `package-lock.json` can go

client/src/views/Signup.tsx
esteban-gs: This is documented here: https://mui.com/material-ui/api/modal/#Modal-prop-onClose

client/src/components/BannerSection.tsx
```diff
@@ -44,15 +45,23 @@ function BannerText() {
...
+    }
+  };
   return (
     <Box>
       <Grid container spacing={3} sx={{ height: '600px' }}>
         <Grid xs={6} item justifyContent="center" sx={{ display: 'flex' }}>
           <Box sx={{ width: '550px', mt: '180px', mb: '66px', ml: '40px' }}>
             <BannerText />
-            <CTAHeroButton text="Join Now" onClick={() => history.push('/signup-citizen')} />
+            <CTAHeroButton text="Join Now" onClick={() => handleOpenModal('SignUp')} />
```
esteban-gs, line 64: 👍🏾

client/src/components/Buttons/CTAButton.tsx
esteban-gs, line 1: @denniswangcodes , this should already be implemented on the main branch. See: https://github.com/Nonprofit-Exchange-Hub/web-app/blob/14e7f562f77786f4e04d14f39c81460bb78ef9c5/client/src/components/Buttons/Button.tsx#L38

client/src/components/BannerSection.tsx
```diff
@@ -44,15 +45,23 @@ function BannerText() {
 
 function BannerSection() {
   const { classes } = useStyles();
-  const history = useHistory();
+  const modalContext = useContext(ModalContext);
+  const { openModal } = modalContext;
```
esteban-gs, line 49: I think this will also work for this
```suggestion
  const { openModal } = useContext(ModalContext);
```

client/src/components/Buttons/CTAButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+}));
+
+type Props = {
+  text: string;
+};
+
+function CTAButton({ text }: Props) {
+  const { classes } = useStyles();
+  const modalContext = useContext(ModalContext);
+  const { openModal } = modalContext;
```
esteban-gs, line 29:
```suggestion
    const { openModal } = useContext(ModalContext);
```

client/src/components/Icons/SvgSignUpProfileStep.js
```diff
@@ -0,0 +1,1513 @@
+import * as React from 'react';
```
esteban-gs, line 1: Do we need to import * from React here?

client/src/components/Icons/SvgSignUpProfileStep.js
esteban-gs, line 1: It'll probably be better to add these to the `assets` folder and save them as `svg` instead of JS. You can export these from figma.
Then, you can import it like this: https://github.com/Nonprofit-Exchange-Hub/web-app/blob/d1b357a877629486c8af978bda50cea49bbd2e9e/client/src/components/FAQs.tsx#L6

esteban-gs (approved):
Approved once comments are addressed.

Issue #311, opened 2023-03-31
home page how it works section


jd2rogers2:
restyle the "how it works" part of the home page (found in HowItWorks.tsx) to match the mocks (screen shot below)

find the static assets in the google drive (if they aren't there yet ping ally and or use something similar) https://drive.google.com/drive/folders/17TyKrqROm9epjrv_cQvvQwNmniNInkq5?usp=share_link

![Image](https://user-images.githubusercontent.com/16400333/228995280-cb541200-7c77-455d-88ba-3304697ec06a.png)

PR #349, opened 2023-06-12
Home page how it works section #311

taoliu12:
How it works card section completed.

## Test Plan

repro steps:

1. localhost:3000 
2. Be able to view "how it works section" on homepage


client/src/views/HowItWorks.tsx
```diff
@@ -1,133 +1,98 @@
...
-
-const nonprofitInstructionList = [
+const nonprofitinstructionsArray = [
   {
-    title: '1. Post your Need.',
-    body: loremIpsum,
-    image: 'https://picsum.photos/seed/nonprofit/600',
+    title: 'Sign up with Givingful',
+    body: 'Create a profile to let our community know about your nonprofit.',
+    image: HowItworks1,
```
esteban-gs, line 19: I think we could enhance readability by renaming all these images to what they do, for example: `sign-up-with-givingful.svg`

client/src/views/HowItWorksCards.tsx
esteban-gs, line 1: Both this and the `HowItWorksCard` component can go in the components directory. We are reserving `src/views` for top-level page views.

esteban-gs (changes requested):
@taoliu12 , thanks for working on this. I just have a couple of recommendations.

Also, the build is failing, but not due to your code. I think if we drop the `--legacy-peer-deps` from the line [here](https://github.com/Nonprofit-Exchange-Hub/web-app/blob/318a504a5f4bea6cdeef6ecaca63c6b0e795f443/Dockerfile#L8), it will build successfully.

client/src/views/HowItWorksCard.tsx
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable prettier/prettier */
```
esteban-gs, line 1: @taoliu12 , it seems like this eslint comment is causing the build to break.
```suggestion

```

client/src/views/HowItWorks.tsx
```diff
@@ -1,133 +1,98 @@
...
 import { makeStyles } from '@mui/styles';
-import { Box, Grid, Container } from '@mui/material';
+import { Box, Grid } from '@mui/material';
 
 import type { Theme } from '@mui/material/styles';
+import HowItWorksCards from './HowItWorksCards';
+import PeopleSignup from '../assets/HowItWorks/people-signup.svg';
+import PeopleSharing from '../assets/HowItWorks/people-sharing.svg';
+import PeopleWorking from '../assets/HowItWorks/people-working.svg';
+import PeopleFlag from '../assets/HowItWorks/people-flag.svg';
```
taoliu12, line 11: these are the new SVG names

esteban-gs (approved):
Looks good to me!

PR #355, opened 2023-07-17
Fix profile page

rmwerner18:
-Show interests, bio, city/state/zip, and profile image on profile
-add 'Edit Profile' button


esteban-gs (approved):
LGTM!

esteban-gs (changes requested):
Actually, I just found some issues that we'll need to address.

esteban-gs (approved):

PR #362, opened 2023-07-27
updated the footer with the mockup provided

EtoKruto:
still need confirmation on correct navlinks


EtoKruto:
@marvinsjsu, can you confirm if the links used are correct or if we should link to the 'Our Mission" page, which hasn't been created yet? I could not find it in the mockups, so I want to confirm.

marvinsjsu:
@EtoKruto, this all looks great.  I think the links are going to the right pages, if not, we'll create a new ticket for any needed update.  I'm not sure if you were able to get the padding/margin to help determine spacing of the elements (I wasn't able to), so I've asked UI/UX if they can help with this.  The only visual comment I have is with regards to the elevation (the box-shadowing) of the footer itself - it can be adjusted to match closer to what we have on the mocks.  Other than that, I'll go over the code sometime this evening, but functionally it looks great!

Screenshot 2023-07-27 at 12 51 18 PM

EtoKruto:
@marvinsjsu should be good to go for the review. Thanks!

marvinsjsu (approved):
LGTM and thanks for the tip on finding/displaying the spacing on Figma!

PR #368, opened 2023-08-11
react-error-boundary commit

EtoKruto:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


marvinsjsu:
hi @EtoKruto, I was reading up on `ErrorBoundary` and I think you've set everything up correctly, but we're using an event handler for triggering the error, which looks like `ErrorBoundary` won't be able to catch (based on https://medium.com/@danhuang1202/catch-error-from-event-handler-in-react-error-boundary-f36ec58786af).  I've tried locally by just throwing the error in `SignInModal` by doing something like ...
```
if (!isLoading) {
  throw new Error('An intentional error');
}
```

This will show us the error pane, which I think will always show for development (will be the `ErrorBoundary` when production), but we can close it and we'll see the `ErrorBoundary` displayed.

Screenshot 2023-08-31 at 8 46 01 PM

Screenshot 2023-08-31 at 8 46 13 PM

marvinsjsu:
I couldn't find this mentioned in the new React dev docs, but they do have it in the legacy site: 
- https://legacy.reactjs.org/docs/error-boundaries.html#introducing-error-boundaries
- https://legacy.reactjs.org/docs/error-boundaries.html#how-about-event-handlers

EtoKruto:
@marvinsjsu please feel free to review when you find some time. Thank you sir!

client/src/App.tsx
```diff
@@ -29,30 +25,37 @@ function App(): JSX.Element {
...
-          </UserProvider>
+                  <Grid
+                    className="App"
+                    display="flex"
+                    flexDirection="column"
+                    height="100vh"
+                    width="100vw"
+                    justifyContent="space-between"
+                  >
+                    {/* move modal to below footer after troubleshooting */}
```
marvinsjsu, line 49: I'm assuming this comment was added before you started working on this, but do you know what this comment is about?

client/src/App.tsx
```diff
@@ -29,30 +25,37 @@ function App(): JSX.Element {
...
+                    <Modal />
+                    <Header />
+                    <Main />
```
marvinsjsu, line 52: It might also be good to wrap `Main` around with an `ErrorBoundary`.  This is so when an error is inside the main component, we will still have a header and footer.

Screenshot 2023-09-11 at 6 25 28 AM

marvinsjsu: Let's reach out to design team and see if they can help create a display for these kinds of errors (including 404 - page not found) .. maybe they can have some fun with it like in some of these https://www.creativebloq.com/web-design/best-404-pages-812505 ... which reminds me, I need to do the same with the loading and error states for the user sign-up form (will drop a message on the design channel later today).

marvinsjsu (approved):
Aside from a suggestion and a question, this LGTM.  Thanks for workin on this, @EtoKruto!

client/src/App.tsx
EtoKruto: Truly not sure, might need to go a few commits back to see the original placer of the said Modal position. I created the Modal and modal provider originally but do not remember adding the comment. I've since moved it to below footer for order but to my best understanding this doesn't directly impact its appearance or position on the screen, especially since it's styled with absolute positioning

client/src/App.tsx
EtoKruto: 404 Pages are always the tell if the devs cared or not (or if they had plenty of time or were rushed perhaps?). Either way i like the idea!

Moving the ErrorBoundary around the Main makes sense for now unless we need a greater scope, which should only require a simple move once again.

EtoKruto:
@marvinsjsu Please provide a quick glance on the new structure if possible - I've folderized the ErrorBound so its compartmentalized into it's own folder for additions in the future and so it's easy on the eyes.

client/src/App.tsx
```diff
@@ -29,30 +26,33 @@ function App(): JSX.Element {
...
+                  <Grid
+                    className="App"
+                    display="flex"
+                    flexDirection="column"
+                    height="100vh"
+                    width="100vw"
+                    justifyContent="space-between"
+                  >
+                    <Header />
+                    <ErrorBoundary fallback={NotFound}>
```
marvinsjsu, line 46: I think the `fallback` prop should be just like what's in the outer `ErrorBoundary`, being `GeneralError`.  I think I may have confused things when I mentioned that we're not handling 404/page not found error and that we need to reach out to design for this.  I think for this kind of error, instead of using the `ErrorBoundary`, we'd need to add a route that basically shows a 404 page.  I've reached out to design on [#design-ui-ux channel](https://nonprofitcircle.slack.com/archives/C01G4GR05CJ/p1694800626038909) regarding this.

marvinsjsu (approved):
I like how you re-organized the components.  It was not required, but it is very helpful as it helps keep a nice structure to our app.  Thanks again, @EtoKruto!

PR #373, opened 2023-09-12
136/sanitize returned api data

esteban-gs:
## Dev Summary

Changes:
- add class-transform serialization interceptor. This will use the `class-transformer` annotations to explicitly expose what we want to expose on data going out.
- sanitize return user-dto vals. Apply the transform interceptor to the account controller `register` endpoint.

## Test Plan

repro steps:

See the the test that verifies the password isn't returned: https://github.com/Nonprofit-Exchange-Hub/web-app/blob/50b1240971b39ceb5faa5f8a12ab23ab82d3dfb6/server/test/account-manager/account-manager.controller.e2e-spec.ts#L87

## Resources

- https://github.com/typestack/class-transformer

### DIFF
--- a/server/src/acccount-manager/dto/auth.dto.ts
+++ b/server/src/acccount-manager/dto/auth.dto.ts
@@ -1,6 +1,9 @@
-import { IsEmail } from 'class-validator';
+import { IsEmail, IsNotEmpty, IsString } from 'class-validator';
+import { Expose } from 'class-transformer';
 
 export class VerifyEmailDto {
+  @IsString()
+  @IsNotEmpty()
   token: string;
 }
 
@@ -16,18 +19,30 @@ export class ResetPasswordDto {
 }
 
 export class ReturnUserDto {
+  @Expose()
   id: number;
+  @Expose()
   firstName: string;
-  last_name: string;
+  @Expose()
+  last_name?: string;
+  @Expose()
   email: string;
+  @Expose()
   bio?: string;
-  city: string;
-  state: string;
+  @Expose()
+  city?: string;
+  @Expose()
+  state?: string;
+  @Expose()
   zip_code: string;
+  @Expose()
   email_notification_opt_out: boolean;
+  @Expose()
   email_verified: boolean;
+  @Expose()
   profile_image_url?: string;
 }
+
 export class ReturnSessionDto {
   user: ReturnUserDto;
 }
--- a/server/src/acccount-manager/user.service.ts
+++ b/server/src/acccount-manager/user.service.ts
@@ -2,7 +2,6 @@ import { HttpException, HttpStatus, Injectable, Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import * as bcrypt from 'bcryptjs';
 import { Repository } from 'typeorm/repository/Repository';
-import { ReturnUserDto } from './dto/auth.dto';
 import { CreateUserInternal, UpdateUserInternal } from './dto/create-user.internal';
 import { User } from './entities/user.entity';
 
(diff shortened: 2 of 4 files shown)


server/src/acccount-manager/dto/auth.dto.ts
```diff
@@ -16,18 +19,30 @@ export class ResetPasswordDto {
 }
 
 export class ReturnUserDto {
+  @Expose()
   id: number;
+  @Expose()
   firstName: string;
-  last_name: string;
+  @Expose()
+  last_name?: string;
```
marvinsjsu, line 27: Do you think it's worth to add a migration to update field names to follow a naming convention with camel-case instead of snake-case, or vice-versa?  I think my preference is camelCase as it aligns with the front-end, but I'm not sure what is more common with Node/Nest.js?

server/src/shared/serialize.interceptor.ts
```diff
@@ -0,0 +1,25 @@
+import { CallHandler, ExecutionContext, NestInterceptor, UseInterceptors } from '@nestjs/common';
+import { Observable, map } from 'rxjs';
+import { plainToInstance } from 'class-transformer';
+
+export const MapTo = (dto: any) => {
+  return UseInterceptors(new SerializeInterceptor(dto));
+};
+
+export class SerializeInterceptor implements NestInterceptor {
```
marvinsjsu, line 9: Is an interceptor similar to a middleware? (I should go to the link and read up on `class-transformer` :D)

marvinsjsu: In `class-transformer` [documentation](https://github.com/typestack/class-transformer#nodejs), it looks like we may also need `reflect-metadata` and don't see it as a dependency.

server/src/acccount-manager/user.service.ts
```diff
@@ -12,12 +11,11 @@ const { BCRYPT_WORK_FACTOR = '10' } = process.env;
 export class UsersService {
   constructor(@InjectRepository(User) private usersRepository: Repository<User>) {}
 
-  async create(createUserDto: CreateUserInternal): Promise<ReturnUserDto> {
+  async create(createUserDto: CreateUserInternal): Promise<User> {
     try {
       const hashedPw = await bcrypt.hash(createUserDto.password, parseInt(BCRYPT_WORK_FACTOR));
       createUserDto.password = hashedPw;
       const user = await this.usersRepository.save(createUserDto);
-      delete user.password;
```
marvinsjsu, line 20: I've verified locally that `password` is not included in the returned payload even with this line removed.  I'm not 100% clear on how this is happening, so describing here what I think is going on.  In `UsersService.create` method, we've replaced `ReturnUserDto` as the output type with `User`, but in the controller, we use annotation `@MapTo(ReturnUserDto)`, which I'm assuming is doing the sanitizing.   Since `password` is not one of the properties of `ReturnUserDto`, what does the `@Expose()` annotation do and why do we use it?

marvinsjsu (approved):
As mentioned in a previous comment, I've verified that the `password` data is not in the response and sanitizing looks to be working.  I just have a few questions, but this LGTM.

server/src/acccount-manager/dto/auth.dto.ts
esteban-gs: That's a good point. This was done, I believe to make table column names conform to the Postgres standard, but there should be a way to tell Typeorm to use camel case in JS.

server/src/shared/serialize.interceptor.ts
esteban-gs: It's satisfied by the nestjs/common peer dependencies, which uses it as well.

esteban-gs: > Is an interceptor similar to a middleware? 

In NestJs an interceptor is middle-ware. In this specific case, we're chaining a `class-transformer` operation on the outgoing handler response. It's applied to any handler (controller method) using the `UseInterceptors(new SerializeInterceptor(dto))` decorator, but in our case, I just wrapped it in a shorter decorator `MapTo`.

server/src/acccount-manager/user.service.ts
esteban-gs: Yes, in a way. changing the return type actually has no bearing on the returned data (thanks for Javascript). Thankfully, `class-transformer` is very good at this. The class-transformer interceptor is getting set to
```typescript
 return plainToInstance(this.dto, data, {
          excludeExtraneousValues: true,
          strategy: 'excludeAll',
        });
```

The `excludeAll` strategy means that any `ReturnDto` property not decorated with `Expose` will be removed from the response.

esteban-gs: I went with `excludeAll` so that we have to explicitly list all the data we want to expose, but we can change it if it becomes to cumbersome.

PR #374, opened 2023-09-14
modified header to look like mock

EtoKruto:
'modified header to look like mock - used user, if defined, to show icons: inbox, notification and profile'

### DIFF
--- a/client/src/components/Header.tsx
+++ b/client/src/components/Header.tsx
@@ -21,8 +21,7 @@ import Logout from '@mui/icons-material/Logout';
 import NotificationsNoneIcon from '@mui/icons-material/NotificationsNone';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
-
-import type { Theme } from '@mui/material/styles';
+import theme from '../theme';
 
 import { UserContext } from '../providers';
 import Logo from '../assets/GivingfulLogo.png';
@@ -31,7 +30,7 @@ import { APP_API_BASE_URL } from '../configs';
 import { UserAvatar } from './Users/UserAvatar';
 import { ModalContext } from './../providers/ModalProvider';
 
-const useStyles = makeStyles()((theme: Theme) => ({
+const useStyles = makeStyles()(() => ({
   home: {
     minWidth: '25px',
     maxWidth: '100px',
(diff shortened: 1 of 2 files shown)


client/src/components/Header.tsx
```diff
@@ -95,6 +96,17 @@ const useStyles = makeStyles()((theme: Theme) => ({
     marginLeft: '10px',
     width: '100px',
   },
+  menuButton: {
+    color: '#323232',
+    fontFamily: 'Poppins',
```
esteban-gs, line 101: Font family is set at the theme level, so, you could remove it.
```suggestion
```

client/src/components/Header.tsx
```diff
@@ -95,6 +96,17 @@ const useStyles = makeStyles()((theme: Theme) => ({
...
+    fontSize: '1rem',
+    fontStyle: 'normal',
+    fontWeight: 400,
+    lineHeight: 'normal',
+    letterSpacing: '0.00938rem',
+    width: '100%',
+    margin: '0 5px 0 5px',
```
esteban-gs, line 108: Please check if any of these aren't already set at the theme level.

client/src/components/Header.tsx
```diff
@@ -171,164 +183,189 @@ function Header() {
...
+                id="exchange-button"
+                aria-label="exchange dropdown"
+                aria-controls={isExchangeMenuOpen ? 'about-menu' : undefined}
+                aria-haspopup="true"
+                aria-expanded={isExchangeMenuOpen ? 'true' : undefined}
+                onClick={handleClick}
+                endIcon={isExchangeMenuOpen ? <ExpandLessIcon /> : <ExpandMoreIcon />}
+                sx={{
+                  textTransform: 'capitalize',
+                  color: '#323232',
```
esteban-gs, line 201: It'll be more maintainable to get the primary/secondary color from the theme definition.

`theme.palette.primary.main`

Theme is imported from `src/theme/index.tsx`

esteban-gs (approved):
@alexislcstone , Thanks for working on this! I added a couple of comments for improvements.

I think we could also benefit from splitting up `Header` into smaller components, but up to you if want to tackle it in this PR.

EtoKruto:
> @alexislcstone , Thanks for working on this! I added a couple of comments for improvements.
> 
> I think we could also benefit from splitting up `Header` into smaller components, but up to you if want to tackle it in this...

Totally think it is a good idea but rather leave it for another sprint or for someone else to split, there is some refactoring too I can see. @esteban-gs I addressed the questions and will be pushing to main here shortly.

PR #375, opened 2023-09-16
364/submit profile image

esteban-gs:
## Dev Summary

- Standardize CTA buttons per our design docs
- Use new standard CTA buttons
- Make step four form use form input element
- New profile type
- Update user profile after register (this won't fully work until we fix out email verification)

## Test Plan

repro steps:

1. go to `signup-citizen`
2. fill out the form, add an image in the `upload` part of the 4th step. Expect to see validation on extension and file size
3. watch the network call
4. For now, this causes an unauthorized response because the user hasn't yet confirmed their email


client/src/components/Buttons/Button.tsx
esteban-gs, line 1: I'm making this file the space where we put all the button variants in our designs: https://www.figma.com/file/KE2Sw6ob07agN25RmeNuFg/GiveCycle-wireframes?type=design&node-id=7744-49098&mode=design&t=9mGfginQIeYvCQFq-0

client/src/components/Users/Auth/SignUpCitizen/apis/backend.ts
```diff
@@ -27,6 +27,12 @@ const Endpoints = {
   checkUserEmail: (userEmail: string) => {
     return client.get(`/auth/user-email-exists/${userEmail}`);
   },
+  userUpdateProfile: (profile: Profile) => {
+    const { file, userId } = profile;
+    const formData = new FormData();
+    formData.append('profile_image', file);
```
marvinsjsu, line 33: Could we also use `formData.set` here?  (just for my own understanding, not a request to change)

client/src/components/Users/Auth/SignUpCitizen/StepFour.tsx
```diff
@@ -48,24 +66,24 @@ export default function StepFour({ initData, handleBack, handleNext }: TStepFour
...
+              text="Upload"
+              onChange={handleImageChange}
+              buttonVariant="primaryCTAButton"
+            ></FileUploadInput>
+          </Grid>
+          <Grid item xs={12}>
+            {imageError === 'too-big' ? (
+              <Alert severity="error">Please upload a file up to 1 megabyte in size.</Alert>
+            ) : imageError === 'unsupported-ext' ? (
+              <Alert severity="error">Supported files: jpeg, jpgm png and gif</Alert>
```
marvinsjsu, line 85: Is `jpgm` maybe a typo? (maybe meant to be `jpg`)

client/src/components/Users/Auth/SignUpCitizen/apis/backend.ts
esteban-gs: from Mozilla docs:
>The difference between [set()](https://developer.mozilla.org/en-US/docs/Web/API/FormData/set) and append() is that if the specified key already exists, set() will overwrite all existing values with the new one, whereas append() will append the new value onto the end of the existing set of values.

client/src/components/Users/Auth/SignUpCitizen/StepFour.tsx
esteban-gs: it's a typo! Fixed!

marvinsjsu:
hi @esteban-gs, one thing I noticed is that when first choosing an invalid image format, we get the error message displayed, which is great, but this message continues to be displayed even if subsequent selection for image is valid.  Also, maybe this can be a follow-up ticket, but once a valid image is selected, maybe it'd be nice to show it instead of the default image for profile.

marvinsjsu (approved):
tested locally and LGTM, just added a comment regarding error message and maybe a follow-up ticket

PR #380, opened 2023-10-14
Search issue 361

taoliu12:
## Dev Summary

[Mocks](link)

## Test Plan

repro steps:

1. localhost:3000 
2. click X button
expected: 

## Resources


client/src/views/AssetsList.tsx
```diff
@@ -51,7 +51,7 @@ type Props = {
 function AssetsList(props: Props): JSX.Element {
   const { classes } = useStyles();
   const { assets, headerContentRight, headerText } = props;
-  if (!assets) return <> </>;
+  if (!assets || !Array.isArray(assets)) return <> No Results</>;
```
taoliu12, line 54: assets was coming back as an error object when DB is unseeded, so I added a check for if it is an array.

esteban-gs (changes requested):
This looks fine to me, maybe safer to do a nullish check like:

`{assets && assets.map(...`

Looking into this, it seems that the server returning a 500 error should be handled in the fetch call. Right now that fetch call isn't catching any errors.

https://github.com/Nonprofit-Exchange-Hub/web-app/blob/d1b357a877629486c8af978bda50cea49bbd2e9e/client/src/views/SearchResults.tsx#L88

PR #385, opened 2023-11-06
377/ui categories from db

esteban-gs:
## Dev Summary

- Delete unused code
- App settings provider
- Use SettingsProvider to load categories from backend instead of the hard-coded ones

## Test Plan

repro steps:

1. Start the app and go through the user signup flow
2. With the browser inspector open, see the network traffic.
3. Categories should be fetched from the backend when the app first loads.
4. When you get to the 3rd step in the signup flow, the categories (user interests) should be visible and be the ones that have been seeded into the database

## Resources

[Devtrium](https://devtrium.com/about)


client/src/components/Users/Auth/SignUpCitizen/constants/interests.ts
esteban-gs, line 1: With this data now, being fetched from the backend, this is no longer needed.

client/src/components/Users/Auth/SignUpCitizen/validation-rules.ts
esteban-gs, line 1: Our new implementations of form validation don't used this anymore

PR #391, opened 2023-11-29
add dynamic data to useeffect dep array

esteban-gs:
## Dev Summary

This is a simple bugfix for a useEffect without the correct dep array.

This caused the SearchResult component to continously render without end.

## Test Plan

repro steps:

1. localhost:3000 
2. search anything
expected: 

The search no longer loops indefinitely

## Resources


unsolvedanomaly (approved):
looks good to me

